PR #15, opened 2018-01-17
Fix binaries not copied to install\bin

hyumo:
Create a directory install\bin before copying


lochel:
Don't delete an existing folder `install\bin`, because it may contain binaries for a different target.

I suggest something like:
```batch
IF NOT EXIST "install\bin" MKDIR "install\bin"
```

PR #17, opened 2018-01-23
Add support for VS 2017

lochel:
Boost 1.63 doesn't fully support VS2017. I am not sure how to deal with it. Maybe we can get remove the dependency to Boost.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,11 @@
-cmake_minimum_required(VERSION 3.1)
+IF(WIN32)
+  # VS 2017 requires Boost 1.64 or newer
+  # Boost 1.64 requires CMake 3.8 or newer
+  cmake_minimum_required(VERSION 3.8)
+ELSE()
+  cmake_minimum_required(VERSION 3.1)
+ENDIF()
+
 project(OMSimulator)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/config.cmake/")
--- a/README.md
+++ b/README.md
@@ -32,13 +32,20 @@ OpenModelica FMI &amp; TLM based simulator
 
 ### Windows
 
-The batch scripts are developed for *MS Visual Studio 14 2015 Win64*.
+The following versions of MS Visual Studio are supported:
+
+- MS Visual Studio 2015 [Win64]:
+  - configWinVS14.bat
+  - buildWinVS14.bat
+- MS Visual Studio 2017 [Win64]:
+  - configWinVS15.bat
+  - buildWinVS15.bat
 
 It is not strictly required to install the full Visual Studio IDE. The batch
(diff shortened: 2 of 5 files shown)


lochel:
Here is a workaround to compile Boost 1.63 with VS 2017: https://social.msdn.microsoft.com/Forums/en-US/185d7034-77ed-4394-b1e9-c8f6fb819034/adding-boost-to-a-vs2017-project?forum=visualstudiogeneral
However, I haven't tried it yet.

hyumo:
It would be great if Boost dependency could be removed. Incrementing Boost version could also be an option. VS2017 prebuilt binaries are available starting from [Boost 1.64](https://sourceforge.net/projects/boost/files/boost-binaries/1.64.0/)

lochel:
@hyumo I cannot find the VS2017 binaries. However, upgrading the boost libraries causes other issues on Linux, as I remember from a discussion with @sjoelund.

hyumo:
@lochel I think msvc-14.1(VS2017) binaries are available in the 1.64 prebuilt list see [32bit](https://sourceforge.net/projects/boost/files/boost-binaries/1.64.0/boost_1_64_0-msvc-14.1-32.exe/download)/[64bit](https://sourceforge.net/projects/boost/files/boost-binaries/1.64.0/boost_1_64_0-msvc-14.1-64.exe/download). Just ignore me if higher boost versions affect Linux compilation. Thanks for your respond.

lochel:
I see; VS 2017 is actually msvc-14.1. However, upgrading boost is unfortunately not an option.

adrpo:
Can you support different boost versions via some ifdefs?

lochel:
The cmake files require boost 1.54 as a minimum. So it should work with different boost versions. However, we had issues with different simulation output for different boost versions in the past, but that could be fixed in the meantime.

sjoelund:
@lochel The issue was that you thought an older Boost produced different results, right? But actually it was because a non-ordered set was used somewhere and it was just luck that made simulations succeed. I don't think we had much problem due to different boost versions being used for this project. The problem with requiring newer Boost versions is that they are not available in Ubuntu  (2014 LTS is still supported and only has Boost 1.54; 17.10 has Boost 1.62 and 18.04 LTS will have Boost 1.65), but of course Windows doesn't come with either of them so you can just recommend in the instructions that you install Boost 1.64+ for the VS2017 binaries.

lochel:
I tried now to compile everything with VS2017 and Boost 1.64, but for some reason cmake cannot find the Boost libs.

lochel:
Apparently, boost 1.64 requires cmake 3.8 or newer.

sjoelund:
You shouldn't require cmake 3.8; the project builds just fine with lower versions (cmake 2.8 is in Ubuntu 14.04 LTS; cmake 3.5 in 16.04 LTS). Most systems with a lower version of cmake will have a lower version of boost anyway (or just have the higher cmake_require for Windows?)

PR #21, opened 2018-01-30
Link always to the static fmilib

lochel:


lochel:
@lenaRB Can you please check that this PR doesn't break the macosx build?

adeas31:
Also need to link with `shlwapi` for mingw otherwise we will get compilation error `undefined reference to __imp_UrlCreateFromPathA`.

lochel:
I am going to rebase this changes in order to solve the conflicts with the current master.

PR #30, opened 2018-02-09
run doStep in parallel

adrpo:
add some stuff to .gitignore
force MSYS makefiles


adrpo:
Not good enough, rejected!

PR #33, opened 2018-02-09
Return status from oms_instantiateFMU

adeas31:
Don't call exit(1) as it would quit OMEdit as well.


lochel:
The new API will not use class `CompositeModel`.

lochel:
I will remove `logFatal` from the logging system. I think there is no point keeping it, if we do not calle `exit()`.

lochel:
I've removed `logFatal` in #34.

PR #38, opened 2018-02-12
Don't print the new temp directory message twice

adeas31:


lochel:
That is actually two different messages. The one you are about to remove is to inform the user that a new directory has been created because the specified temporary directory did not exist.

adeas31:
Oh yeah that's right.

PR #56, opened 2018-02-23
Add option to build with multiple jobs

lochel:
E.g.: `make OMSimulator j=4`


sjoelund:
This shouldn't have been needed; $(MAKE) propagates $(MAKEFLAGS) so make -j4 should have worked and seems like a better choice than a custom variable. This is also how I built OMSimulator in the past, and it did work in parallel then...

lochel:
Really? I agree, that is much better then.

PR #59, opened 2018-02-23
Use reasonable default values for geometry

adeas31:


lochel:
The geometry information is optional and treated as undefined if `y1 == y2`. OMEdit should override it immediately when creating new objects.

adeas31:
Yeah but when OMEdit is reading the value it should get a reasonable default instead of 0.

lochel:
We can make it returning a NULL pointer if actually no information is specified. But I guess OMEdit can handle `y1 == y2` as unspecified though.

adeas31:
So is basically undefined even when `x1 == x2`.

lochel:
I would treat it as undefined if a component's area is zero, i.e. `y1 == y2 || x1 == x2`.

lochel:
Okay, I will make it optional and returning a NULL pointer if undefinied.

PR #64, opened 2018-02-27
New API: oms2_rename()

lochel:
Renames a composite model (works for FMI and TLM) or sub-model (e.g. FMU instance).


lochel:
I forgot to update the connections...

PR #66, opened 2018-02-28
Read and Write the ConnectionGeometry

adeas31:


adeas31:
@lochel There might be better ways to do this so please feel free to update this branch.

lochel:
We can merge it as it is for now.

PR #67, opened 2018-02-28
oms_connection_t: Connection between two connectors

lochel:
- [x] Add geometry to type `oms_connection_t`
- [x] New API `oms2_addConnection()`
- [x] New API `oms2_deleteConnection()`
- [x] New API `oms2_updateConnection()`

### DIFF
--- a/src/OMSimulatorLib/OMSimulator.cpp
+++ b/src/OMSimulatorLib/OMSimulator.cpp
@@ -604,10 +604,10 @@ oms_status_enu_t oms2_getConnections(const char* cref, oms_connection_t*** conne
   return oms2::Scope::getConnections(oms2::ComRef(cref), reinterpret_cast<oms2::Connection***>(connections));
 }
 
-oms_status_enu_t oms2_addConnection(const char* cref, const oms_connection_t* connection)
+oms_status_enu_t oms2_addConnection(const char* cref, const char* conA, const char* conB)
 {
   logTrace();
-  return oms2::Scope::addConnection(oms2::ComRef(cref), reinterpret_cast<const oms2::Connection*>(connection));
+  return oms2::Scope::addConnection(oms2::ComRef(cref), oms2::SignalRef(conA), oms2::SignalRef(conB));
 }
 
 oms_status_enu_t oms2_deleteConnection(const char* cref, const char* conA, const char* conB)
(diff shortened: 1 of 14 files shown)


lochel:
Functions `getConnectionGeometry()` and `setConnectionGeometry()` got removed. The new functions `oms2_getConnections()` and `oms2_updateConnections()` can be used instead.

adeas31:
`oms2_updateConnections`? are you planning to update multiple connections at once?

lochel:
No, that was just a typo.

lochel:
The API for connections will look like this:

```C++
/**
 * \brief Get list of all connections from a given component.
 *
 * \param cref          [in] Full identifier of a component
 * \param connections   [out] Array of connections (null-terminated array)
 * \return              Error status.
 */
oms_status_enu_t oms2_getConnections(const char* cref, oms_connection_t*** connections);

/**
 * \brief Adds a new connection to a given parent component.
 *
 * \param cref         [in] Full identifier of a component
 * \param connection   [in] New connection
 * \return             Error status
 */
oms_status_enu_t oms2_addConnection(const char* cref, const oms_connection_t* connection);

/**
 * \brief Deletes the connection between connectors A and B.
 *
 * \param cref   [in] Full identifier of a component
 * \param conA   [in] Name of connector A
 * \param conB   [in] Name of connector B
 * \return       Error status
 */
oms_status_enu_t oms2_deleteConnection(const char* cref, const char* conA, const char* conB);

/**
 * \brief Updates the connection between connectors A and B.
 *
 * \param cref         [in] Full identifier of a component
 * \param conA         [in] Name of connector A
 * \param conB         [in] Name of connector B
 * \param connection   [in] New connection
 * \return             Error status
 */
oms_status_enu_t oms2_updateConnection(const char* cref, const char* conA, const char* conB, const oms_connection_t* connection);
```

lochel:
I changed `oms2_addConnection()` slightly. It creates now a plain connection without additional graphical information. This is better for the scripting interfaces. One will need to call `oms2_updateConnection()` in order to set graphical information.

adeas31:
`oms_status_enu_t oms2_addConnection(const char* cref, const oms_connection_t* connection)`
It takes `oms_connection_t` and if `ssd_connection_geometry_t` is defined then it should add the graphical information and if its NULL then skip it.

lochel:
No, looks now like this:
```C++
oms_status_enu_t oms2_addConnection(const char* cref, const char* conA, const char* conB);
```

adeas31:
How about optional argument `ssd_connection_geometry_t`?

lochel:
We could do that. But I think it is okay to just call `oms2_addConnection` && `oms2_updateConnection` when creating a new connection in OMEdit and `oms2_updateConnection` when updating an existing connection.

PR #71, opened 2018-03-03
New API oms2_getFMUPath()

adeas31:
Add FMU path to `oms_component_t`

### DIFF
--- a/src/OMSimulatorLib/OMSimulator.cpp
+++ b/src/OMSimulatorLib/OMSimulator.cpp
@@ -604,6 +604,12 @@ oms_status_enu_t oms2_getComponents(const char* cref, oms_component_t*** compone
   return oms2::Scope::getComponents(oms2::ComRef(cref), components);
 }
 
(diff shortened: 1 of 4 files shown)


lochel:
The `oms_component_t` is also used for constant values, ports, and lookup tables. Therefore, I am not sure if we should save the path in the component structure.

adeas31:
In that case the path could be empty. We can anyway distinguish about it from `oms_component_type_enu_t`.

lochel:
Sure, that is possible. But that would probably be inconsistent. Other sub-models have other characteristic attributes. We should either add all the different attributes to `oms_component_t` or none of them. I prefer to have a function to get the fmu path, e.g. `oms2_getFMUPath(const char* cref, const char** path)`.

adeas31:
OK. I am fine with a separate function. Will you add it or should I just update in this same PR?

lochel:
It would be nice if you update this PR.

adeas31:
Should be fine to merge now.

PR #74, opened 2018-03-06
New API: setTempDirectory(), setWorkingDirectory(), setLogFile()

lochel:


lochel:
I just discovered a memory issue. It is probably not connected to this PR, but I will investigate it first anyway.

PR #75, opened 2018-03-07
Don't delete the pointer we reference to

adeas31:


src/OMSimulatorLib/oms2/Element.cpp
```diff
@@ -90,10 +90,11 @@ oms2::Element& oms2::Element::operator=(const oms2::Element& rhs)
   for (int i=0; i<nInterfaces; ++i)
     this->interfaces[i] = reinterpret_cast<oms_connector_t*>(new oms2::Connector(*reinterpret_cast<oms2::Connector*>(rhs.interfaces[i])));
 
+  if (this->geometry)
+    delete reinterpret_cast<oms2::ssd::ElementGeometry*>(this->geometry);
   oms2::ssd::ElementGeometry* geometry_ = new oms2::ssd::ElementGeometry();
   *geometry_ = *reinterpret_cast<oms2::ssd::ElementGeometry*>(rhs.geometry);
   this->geometry = reinterpret_cast<ssd_element_geometry_t*>(geometry_);
-  delete geometry_;
```
lochel: You are right, this need to be fixed. Maybe it is better to do it like this:
```c++
if (this->geometry)
  *reinterpret_cast<oms2::ssd::ElementGeometry*>(this->geometry) = *reinterpret_cast<oms2::ssd::ElementGeometry*>(rhs.geometry);
else
  this->geometry = reinterpret_cast<ssd_element_geometry_t*>(new oms2::ssd::ElementGeometry(reinterpret_cast<oms2::ssd::ElementGeometry>(*rhs.geometry)));
```

src/OMSimulatorLib/oms2/Element.cpp
```diff
@@ -117,7 +118,6 @@ void oms2::Element::setGeometry(const oms2::ssd::ElementGeometry* newGeometry)
   {
     oms2::ssd::ElementGeometry* geometry_ = new oms2::ssd::ElementGeometry(*newGeometry);
     this->geometry = reinterpret_cast<ssd_element_geometry_t*>(geometry_);
-    delete geometry_;
```
lochel: And here the same:

```c++
if (this->geometry)
  *reinterpret_cast<oms2::ssd::ElementGeometry*>(this->geometry) = *reinterpret_cast<oms2::ssd::ElementGeometry*>(newGeometry);
else
  this->geometry = reinterpret_cast<ssd_element_geometry_t*>(new oms2::ssd::ElementGeometry(*newGeometry));
```

src/OMSimulatorLib/oms2/Element.cpp
adeas31: I think I tested it like that but for some reason the following doesn't call the assignment operator,
`  *reinterpret_cast<oms2::ssd::ElementGeometry*>(this->geometry) = *reinterpret_cast<oms2::ssd::ElementGeometry*>(rhs.geometry);`

lochel: Weird. I will test it tomorrow.

lochel: Do you have an example that is calling `oms2::Element::operator=`?

adeas31: No. I expect the above line of code to call it.

lochel: No, that one should call `oms2::ssd::ElementGeometry::operator=`.

adeas31: Ahhh of course. I didn't looked at it closely.
So why do you need an example for `oms2::Element::operator=`?

lochel: I want to test test the changes.

adeas31: I think call `oms2_getElements` and then `oms2_unloadModel` and it crash.

lochel: `oms2::Element::operator=` is not used at all and actually shouldn’t be even implemented.

The crash is caused by `oms2::Element::setGeometry` which can be simplified to:
```c++
void oms2::Element::setGeometry(const oms2::ssd::ElementGeometry* newGeometry)
{
  if (this->geometry)
  {
    delete reinterpret_cast<oms2::ssd::ElementGeometry*>(this->geometry);
    this->geometry = NULL;
  }

  if (newGeometry)
  {
    this->geometry = reinterpret_cast<ssd_element_geometry_t*>(new oms2::ssd::ElementGeometry(*newGeometry));
  }
}
```

lochel: I added a test case for this, which can be activated as soon as this is fixed: https://github.com/OpenModelica/OMSimulator-testsuite/commit/f90bcd5f798499f7ecd232e70b5cd0aa58503472

PR #76, opened 2018-03-07
Point to the allocated char array

adeas31:
`oms2_ComRef` is deleted when `oms2::ComRef::getName()` is finished so calling `const_cast` on it is dangerous and leads to undefined behavior.


lochel:
I fixed the same issue in #74 today :)

PR #85, opened 2018-03-12
Fix linking error for MSVC

sjoelund:
This should fix the linking error I introduced.

### DIFF
--- a/src/OMSimulator/CMakeLists.txt
+++ b/src/OMSimulator/CMakeLists.txt
@@ -9,7 +9,9 @@ find_package(Threads)
 # used for the executable itself.
 
 set(Boost_USE_STATIC_LIBS ON)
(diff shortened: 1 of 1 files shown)


lochel:
@sjoelund Apparently, the Windows build is still broken. Can you have a look?

sjoelund:
The new error seems to be related to the OMFitLib subproject. I didn't touch any file that should influence that.

lochel:
That's weird. It compiles even fine on my system. @bernhard-thiele Can you have a look?

sjoelund:
It could have to do with not running the ceres dependencies, but in that case the MSVC job simply needs to compile that (or properly disable CERES in the cmake project for MSVC).

lochel:
I've build the project with OMFit.

PR #93, opened 2018-03-20
Make OMFit optional for the VS build

lochel:


lochel:
I still think we should get rid of the `CERES` flag and derive it from the `OMFIT` flag. But that could be done in a different PR.

PR #103, opened 2018-04-05
Simulation functionality for external TLM models

robbr48:
- [x] Setting TLM socket data
- [x] Adding TLM connections
- [x] Initializing TLM simulations
- [x] Running TLM simulations with external models


OpenModelica-Hudson:

PR #107, opened 2018-04-06
Export simulation results with new API

lochel:


OpenModelica-Hudson:

PR #116, opened 2018-04-16
API function simulate_asynchronous for FMUs

bernhard-thiele:
The PR contains an implementation for the `simulate_asynchronous` API function for FMI composites. The TLM models are not supported (yet) and only the stubs are provided.

There is currently no test that exercises this function. Since this API call is introduced for OMEditor, there is no Lua or Python scripting interface for `simulate_asynchronous`. Should there be also a scripting interface for this function?


bernhard-thiele:
@adeas31 I wonder if my `simulate_asynchronous` implementation does about the thing that you expect it to? 
I'm thinking of merging it if there is no obvious flaw. What is not in this PR is the implementation of `simulate_asynchronous` for TLM. I had a look at the TLM code but it is not obvious to me how the callback function should be served. Maybe @robbr48 has an idea?

OpenModelica-Hudson:
  - OpenModelica/OMSimulator#116

OpenModelica-Hudson:

PR #127, opened 2018-04-24
Added missing getXXX/setXXX oms2 API functions

bernhard-thiele:


OpenModelica-Hudson:
  - OpenModelica/OMSimulator#127

bernhard-thiele:
@lochel I mechanically added the missing API function using the existing functions as example. A drawback is that there is a lot of code bloat due to cloned pieces of code. Probably one could extract common parts of the setXXX/getXXX functions and reduce the bloat.

lochel:
Yes, that might be a good idea. But I think for now the changes look fine and can be merged.

OpenModelica-Hudson:
  - OpenModelica/OMSimulator#127

OpenModelica-Hudson:

PR #140, opened 2018-05-04
Added argument for specifying callback updating period

bernhard-thiele:
Added argument cbPeriod to oms2_simulate_asynchronous function.
Allows to specify the period duration in which the provided callback
function is called (wall-clock time).

@adeas31


OpenModelica-Hudson:
  - OpenModelica/OMSimulator#140

OpenModelica-Hudson:

adeas31:
I will test it on Monday and let you know.
On a second thought shouldn't the callback be called after every step instead of fixed time interval?

PR #148, opened 2018-05-08
Revised oms2_simulate_asynchronous(..)

bernhard-thiele:
@adeas31 I revised the function as discussed. The callback function is called after every completed step of the composite model. I now also check the return values of the `doStep(..)` function calls to the sub-models and return the "worst" of that values in the callback function. However, that is currently pointless since the `doStep(..)` functions of the sub-models always return `fmi2_status_ok` (hard-coded) instead of propagating back the actual return value of the FMU's "doStep" (`fmi2_import_do_step(..)`).

@lochel I think `oms2::FMUWrapper::doStep(..)` should be changed so that it propagates back the "worst" return value of `fmi2_import_do_step(..)`.

### DIFF
--- a/src/OMSimulatorLib/OMSimulator.cpp
+++ b/src/OMSimulatorLib/OMSimulator.cpp
@@ -666,10 +666,10 @@ oms_status_enu_t oms2_terminate(const char* ident)
   return oms2::Scope::GetInstance().terminate(oms2::ComRef(ident));
 }
 
-oms_status_enu_t oms2_simulate_asynchronous(const char* ident, double cbPeriod, void (*cb)(const char* ident, double time, oms_status_enu_t status))
+oms_status_enu_t oms2_simulate_asynchronous(const char* ident, void (*cb)(const char* ident, double time, oms_status_enu_t status))
 {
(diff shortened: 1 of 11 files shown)


OpenModelica-Hudson:
  - OpenModelica/OMSimulator#148

OpenModelica-Hudson:
**[ASan:](https://test.openmodelica.org/hudson/job/OMSimulator-PR/259/)** The test suite is unstable. :disappointed:

bernhard-thiele:
Uups, I didn't change anything that to my understanding should have any effect on ASan?

lochel:
@bernhard-thiele The tests are not failing due to your changes. These are issues in the current implementation that still need to be fixed. There are 2 OMSysIdent tests and 6 OMTLMSimulator tests failing. Maybe some of the issues are actually inside FMUs.
You can disable these tests for now by specifying `-fast` in the field `ASAN_FLAGS`.

OpenModelica-Hudson:

bernhard-thiele:
@lochel Thanks, worked smoothly :-)

PR #179, opened 2018-05-18
Fix for linking OMSysIdent into OMSimulatorLib

bernhard-thiele:


sjoelund:
Why not just mark the functions as extern? Though I suppose you don't want to ship the static library anyway, so it doesn't matter?

bernhard-thiele:
No I think it doesn't make sense any more to ship the static lib. I already thought of getting completely rid of it. The only purpose that is left for it, is to be linked into the OMSimulator executable.

I'm not sure what you mean with marking the functions as extern? They are marked as `extern "C"` in OMSysIdent.h, but I guess that is not what you mean.

OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/316//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#179
  - OpenModelica/OMSimulator-testsuite#57

OpenModelica-Hudson:
All tests were successful. :smiley:

PR #190, opened 2018-05-28
Terminate the running simulation

### LINKED ISSUE #152: Need an API to cancel the running the simulation
Right now it is possible to start the simulation using `oms2_simulate_asynchronous`. However, there is no way to cancel a running simulation.

adeas31:
Fixes #152 

## Purpose

An asynchronous simulation can be started using `oms2_simulate_asynchronous`. An API to cancel the running the simulation is required.

## Approach

There is no standard way to terminate a detached thread. The only way is to store the thread handles and then use the OS native techniques to kill the thread using the handle.

Use the int pointer as a flag to terminate the simulation. Set the pointer value to 1 and pass it to `oms2_simulate_asynchronous`. Set to 0 when you want to terminate the simulation.

I personally think we don't need to terminate or kill the thread via a new API. Just use the flag to get out of the simulation loop and let the thread finish.

## Type of Change

## Checklist


lochel:
I think we have to come up with a new implementation for oms3. Should we close this PR?

adeas31:
I agree.

PR #229, opened 2018-06-08
Add a Jenkins pipeline

sjoelund:
## Purpose

Started work on making a Jenkins pipeline for the OMSimulator workflow, to replace Hudson.

## Approach

By adding a Jenkinsfile that performs a parallel build with optional runs of test suites depending on if it is a branch, tag, or change request.

The documentation is always generated, but only uploaded for branches and tags, to location:
https://openmodelica.org/doc/OMSimulator/BRANCH_OR_TAG_NAME/

## Type of Change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

---

## Open Questions and Pre-Merge TODOs

- [x] Run the test suites
- [x] Add cached runs of the test suites (to run the most expensive jobs first)
- [x] Add a Windows slave to Jenkins
- [x] Add an ARM slave to Jenkins
- [x] Create a 32-bit Linux docker image for testing that?
- [x] Should builds/tests run only on pull requests or always? This might mess up building documentation for tags...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,428 @@
+pipeline {
+  agent none
+  environment {
+    CACHE_DIR = "testsuite/cache/${env.CHANGE_TARGET ?: env.GIT_BRANCH}"
+  }
+  options {
+    newContainerPerStage()
+  }
+  stages {
+    stage('build') {
+      parallel {
+        stage('linux64') {
+          agent {
+            dockerfile {
+              additionalBuildArgs '--pull'
+              dir '.CI/cache'
+              /* The cache Dockerfile makes /cache/runtest, etc world writable
+               * This is necessary because we run the docker image as a user and need to
+               * be able to have a global caching of the omlibrary parts and the runtest database.
+               * Note that the database is stored in a volume on a per-node basis, so the first time
+               * the tests run on a particular node, they might execute slightly slower
+               */
+              label 'linux'
+              args "--mount type=volume,source=runtest-omsimulator-cache-linux64,target=/cache/runtest"
+            }
+          }
+          environment {
+            RUNTESTDB = "/cache/runtest/"
+            NPROC = "${numPhysicalCPU}"
+          }
+          steps {
+            buildOMS()
+            sh '''
+            # No so-files should ever exist in a bin/ folder
+            ! ls install/linux/bin/*.so 1> /dev/null 2>&1
+            (cd install/linux && tar czf "../../OMSimulator-linux-amd64-`git describe`.tar.gz" *)
+            '''
+
+            archiveArtifacts "OMSimulator-linux-amd64-*.tar.gz"
+
+            partest()
+
+            junit 'testsuite/partest/result.xml'
+
+            // Temporary debugging
+            // archiveArtifacts artifacts: 'testsuite/**/*.log', fingerprint: true
+
+            sh 'make doc doc-html doc-doxygen'
+            sh '(cd install/linux/doc && zip -r "../../../OMSimulator-doc-`git describe`.zip" *)'
+            archiveArtifacts artifacts: 'OMSimulator-doc*.zip', fingerprint: true
+            stash name: 'docs', includes: "install/linux/doc/**"
+          }
+        }
+        stage('linux64-asan') {
+          stages {
+            stage('build') {
+              agent {
+                docker {
+                  image 'docker.openmodelica.org/build-deps:v1.13'
+                  label 'linux'
+                  alwaysPull true
+                }
+              }
+              environment {
+                RUNTESTDB = "/cache/runtest/"
+                ASAN = "ON"
+              }
+              steps {
+                buildOMS()
+                stash name: 'asan', includes: "install/linux/**"
+              }
+            }
+            stage('test') {
+              agent {
+                dockerfile {
+                  additionalBuildArgs '--pull'
+                  dir '.CI/cache'
+                  /* The cache Dockerfile makes /cache/runtest, etc world writable
+                   * This is necessary because we run the docker image as a user and need to
+                   * be able to have a global caching of the omlibrary parts and the runtest database.
+                   * Note that the database is stored in a volume on a per-node basis, so the first time
+                   * the tests run on a particular node, they might execute slightly slower
+                   */
+                  label 'linux'
+                  args "--mount type=volume,source=runtest-omsimulator-cache-linux64-asan,target=/cache/runtest " +
+                       "--cap-add SYS_PTRACE " + // Needed for ASAN
+                       "--oom-kill-disable -m 1024m --memory-swap 1024m" // Needed for ASAN
+                }
+              }
+              environment {
+                RUNTESTDB = "/cache/runtest/"
+                ASAN = "ON"
+              }
+              steps {
+                unstash name: 'asan'
+                partest()
+                junit 'testsuite/partest/result.xml'
+              }
+            }
+          }
+        }
+        stage('linux32') {
+          agent {
+            dockerfile {
+              additionalBuildArgs '--pull'
+              dir '.CI/cache-32'
+              /* The cache Dockerfile makes /cache/runtest, etc world writable
+               * This is necessary because we run the docker image as a user and need to
+               * be able to have a global caching of the omlibrary parts and the runtest database.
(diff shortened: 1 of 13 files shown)


lochel:
> Should builds/tests run only on pull requests or always? This might mess up building documentation for tags...

I think we have to build the simulator also for the actual commits on the master (and maintenance branches) to get the version string right.

Jenkinsfile
```diff
@@ -0,0 +1,76 @@
...
+              alwaysPull true
+            }
+          }
+          steps {
+            sh '''
+            git fetch --tags
+            make config-3rdParty -j`nproc`
+            make config-OMSimulator -j`nproc`
+            make OMSimulator -j`nproc`
+            ./install/linux/bin/OMSimulator --version
```
lochel: This is not needed, because it is now part of the Makefile.

Jenkinsfile
```diff
@@ -0,0 +1,76 @@
...
+          agent {
+            label 'osx'
+          }
+          steps {
+            sh '''
+            PATH="$PATH:/opt/openmodelica/bin"
+            make config-3rdParty -j`nproc`
+            make config-OMSimulator -j`nproc`
+            make OMSimulator -j`nproc`
+            ./install/mac/bin/OMSimulator --version
```
lochel: Not needed; see above

sjoelund:
> I think we have to build the simulator also for the actual commits on the master (and maintenance branches) to get the version string right.

Yes, but for which targets? If we publish the builds we naturally always need to build them. But should they also be tested again?

lochel:
I think we need to run it at least for all main targets: Linux; Mac; Windows MSVC
But it shouldn't be necessary to test them again.

Jenkinsfile
```diff
@@ -0,0 +1,160 @@
...
+      }
+      steps {
+        sh """
+        if test -d '${env.JENKINS_HOME}/userContent/OMSimulator/testsuite/cache/${env.GIT_BRANCH}'; then
+          mkdir -p testsuite/cache/
+          cp -av '${env.JENKINS_HOME}/userContent/OMSimulator/testsuite/cache/${env.GIT_BRANCH}' testsuite/cache/
+        fi
+        """
+        stash name: 'old-linux64-db', includes: "testsuite/cache/${env.GIT_BRANCH}/linux64", allowEmpty: true
+        stash name: 'old-osx-db', includes: "testsuite/cache/${env.GIT_BRANCH}/linux64", allowEmpty: true
```
lochel: Shouldn't this be "testsuite/cache/${env.GIT_BRANCH}/osx"?

sjoelund: Yes, but there are more problems. I will be ready with the caching of the db-files soon

sjoelund:
Upon further thought, we should always run the tests. Because I cache the runtests.pl database only when updating the master.

Note that there are 15 tests failing. I am not sure why... perhaps it's a problem with running on Ubuntu 18.04?

lochel:
I just found the list of failing tests. How can I access the log files of the failing tests?

configWinVS.bat
```diff
@@ -205,6 +205,9 @@ EXIT /B 0
 :pthread
 ECHO # config pthread
 CD 3rdParty\pthread
+IF NOT EXIST install\\ MKDIR install
+IF NOT EXIST install\\win MKDIR install\\win
+IF NOT EXIST install\\win\\lib MKDIR install\\win\\lib
```
lochel: This should be moved to `3rdParty\pthread\buildWinVS.bat`.

sjoelund: After everything here is working, we can start moving things

Jenkinsfile
```diff
@@ -0,0 +1,380 @@
...
+cd "${env.WORKSPACE}"
+(cd install/win && zip -r "../../OMSimulator-win64-`git describe`.zip" *)
+"""
+
+            bat """
+set BOOST_ROOT=C:\\local\\boost_1_64_0
+set PATH=C:\\bin\\cmake\\bin;%PATH%
+
+IF NOT EXIST build\\ mkdir build
+IF NOT EXIST build\\win\\ mkdir build\\win
```
lochel: Why do you create these directories manually?

sjoelund:
The job builds for all platforms now. We should probably disable tests that are not working on the affected platforms though (or it will keep saying the build in unsuccessful).
The performance seems rather OK. About 3 minutes to build (combined) and test all platforms except MSVC64 which takes around 5 minutes or so (but this should still be acceptable).

Note that I started cross-compiling mingw64 (mingw32 doesn't quite work due to the Ubuntu GCC using a different stack unwinding), arm32 and osx (mostly needed because it took half an hour to build on raspberry pi).

sjoelund:
Oh yeah, there is a testsuite failure (maybe due to running on Ubuntu 18.04?):
```
build / linux64 / HelloWorld_cs_Fit.py – OMSysIdent
```
It should probably be fixed elsewhere

lochel:
Did you also port the ASAN check to the jenkins pipeline?

sjoelund:
I have added ASAN to the job now, but it seems to be extremely slow; should it be disabled by default?

lochel:
ASAN is slow, but should be enabled by default. How slow is extremly slow in this case?

sjoelund:
Compiling CERES with ASAN was so slow that Jenkins aborted the task...

lochel:
I think it is not needed to compile CERES/libxml2 with ASAN. Try just to compile OMSimulator with ASAN.

sjoelund:
It seems the problem was compiling under the special flags needed in docker to enable ASAN. I now compile in a normal docker instance and will try to run the ASAN tests without it.

sjoelund:
It builds now and all tests succeed

lochel:
As @sjoelund pointed out, there are a few tests failing with bad error messages.

@robbr48 can you have a look at the OMTLMSimulator tests? They fail randomly with a segmentation fault. I observed this behaviour already earlier on my computer, but for some reason the tests a working fine on Hudson. 

Besides that, there is also a problem with the OMSysIdent tests. I will look into that myself.

lochel:
By the way, it might well be that the issues are actually reported by ASAN. We had to disable the check for OMTLMSimulator and OMSysIdent, because we couldn't figure out how to fix them. So that might be a good starting point.

sjoelund:
Still @lochel should we merge this now so we can start moving some things to jenkins? (Like the documentation upload, etc)

lochel:
That's fine for me.

lochel:
Do you run the tests also Windows?

sjoelund:
No; it's only built on Windows

lochel:
Okay, but we need to run the tests on Windows as well.

lochel:
@robbr48 It could well be that I fixed the tlm test cases with #328. However, ASAN still reports issues.

sjoelund:
Then add this to a new PR and play with Jenkins. The MinGW job is actually cross-compiled so it cannot really run the tests unless you support wine or stash/unstash to a Windows machine. I have no desire to learn MSVC / cmd.exe testing :)

lochel:
Well, we should not test less when moving to Jenkins...

lochel:
Btw, do you know why `OMSysIdent\HelloWorld_cs_Fit.py` is actually executed when running with `-asan -fast`? All OMSysIdent tests should be skipped when running with `-fast` 😕

sjoelund:
No, but we can move some things first. It's always hard to move everything at once. Besides I won't have any time until spring. So it's good if someone else learns Jenkins (preferably someone who knows Windows). It's actually quite easy to do things.

sjoelund:
HelloWorld_cs_Fit.py is only run for linux64 and osxcross. It does not seem to run for linux64-asan

lochel:
Okay, and why is linux64-asan then failing?

sjoelund:
Jenkins thing. If the script (runtests.pl) doesn't output anything for a long time, it thinks the job died (I should really increase that deadline...)

OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/584//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#229

OpenModelica-Hudson:

Issue #169, opened 2018-05-15
Logging and plotting of data


lochel:
- [x] Selection of variables to store and plot
- [x] Subsample simulation for logging
- [x] Split up the result files into several smaller ones instead of one large when simulating sequentially
- [x] Allow to skip writing a result file
- [x] Make buffer size adjustable

meek1:
Needed in Release v2.0.

meek1:
Any examples in the test suite or in documentation on how to use setVariableFilter?

lochel:
Not yet, but I will add one as soon as this feature is available.

meek1:
OK, in the meantime, is it possible to simulate sequentially to split up the result files into several smaller ones instead of one large? I tried the following but it resulted in only one result file named "result1.mat":

-- simulation settings
oms2_setCommunicationInterval("AVSDemo", 1e-3)
-- *** 1st simulation sequence ***
oms2_setResultFile("AVSDemo", "result1.mat")
oms2_setStopTime("AVSDemo", 0.02)
oms2_initialize("AVSDemo")
oms2_simulate("AVSDemo")

-- *** 2nd simulation sequence ***
oms2_setResultFile("AVSDemo", "result2.mat")
oms2_setStartTime("AVSDemo", 0.02)
oms2_setStopTime("AVSDemo", 0.04)
oms2_simulate("AVSDemo")

meek1:
A really nice feature would be to select which variable NOT to store, e.g. all concealed* for black-box FMUs. For several test FMUs, the majority of variables are concealed*.

meek1:
@lochel, when can we expect a first functionality to specify plot interval/down sampling for FMI composite models? It is very needed. For the test cases I'm running, I need a communication interval of approx. 1e-3s for stability reasons but I'm only interested in results data of say 1Hz.

I don't want the spcified plot interval to force the master to take exact steps of certain lenght, rather data should be interpolated from available points (or as a first version take the point closest to the specified if this is significantly less implementation work).

lochel:
@meek1 #237 will provide a new API to specify a logging interval. This approach doesn't use any interpolation. It just skips data point which are too close in time. It works already, but I have to test it a bit more before merging (probably later today).

lochel:
Usage of the new functioanlity: `oms2_setLoggingInterval("model", value)`
```
value E (-inf; 0):                  logging disabled
value E [0; communicationInterval]: logging of all data
[...]

PR #236, opened 2018-06-11
Fix simulating w/o generating a result file

lochel:
### Related Issues

#169

### Purpose

Fix bug when providing an empty result filename.

### Approach

Provide a dummy result writer, that is actually doing nothing.

### Type of Change

### Checklist


lochel:
Maybe we should pass the result writer as pointer instead of reference, to avoid having the dummy result writer.

OpenModelica-Hudson:

PR #271, opened 2018-06-29
simulateTLM() function in Model class

robbr48:
### Purpose

Instead of first obtaining variables from Model class and then calling simulateTLM() in FMICompositeModel class, there shall be a simulateTLM() function in Model class.

### Type of Change

### Checklist


lochel:
Actually, I am thinking about refactoring the code and to remove the `Model` class completely. It seems that FMI and TLM models are quite different.

PR #274, opened 2018-07-14
ignore error if we cannot find git

adrpo:
### Purpose

During Linux nightly builds while compiling OMSimulator we don't have git available but is used in the Makefile to clean libxml2. 

### Approach

Just ignore the error if git is not found.

### Type of Change

- build change

### Checklist


lochel:
Why is there no git available? I guess that means that the OMSimulator build of the nightly builds cannot return its version string...

adrpo:
I will not merge this for now as I installed git in the chroot VMs, we'll see how it goes.

adrpo:
Damn, we actually need this as is a source build and the OMSimulator is actually not a git repository but a source one!

OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/482//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#274

OpenModelica-Hudson:

Issue #278, opened 2018-07-28
problem with regex when using GCC 4.8


hansVIE:
### Description

obviously there is a problem with regex and GCC4.8.   
so  **>>OMSimulator  --version**  when using GCC4.8.5 on LINUX platform x86_64  crashed. 

see also: [](https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions)

### workaround
replacing  of **std::regex**   by   **boost::regex** in _Options.cpp, Options.h, ComRef.cpp and FMUWrapper.cpp_
also included  _-lboost_regex_ and everything  is fine (with some additional modifications in cmake) 

### Version and OS
- Version: OMSimulator v2.0.0-dev-519-g1dabf64-linux

- OS: uname -a
Linux kiste.all 4.4.140-62-default #1 SMP Tue Jul 17 12:14:34 UTC 2018 (951943d) x86_64 x86_64 x86_64 GNU/Linux
distribution: OpenSuse 42.3

lochel:
std::regex is not fully implemented in gcc 4.8. Can you use a different gcc version instead?

hansVIE:
Thanks for feedback.
I'm not an expert in C-programming (more on MODELICA) and so was using an old GCC version provided by the distribution. Presently I have a working OMSimulator available, and before compile the next version i'll update GCC.
--> No more action needed from you (only check compiler version in cmake) 
(--> it's not a BUG )

PR #279, opened 2018-08-02
Test for minimum compiler version

lochel:
### Related Issues

#278

### Purpose

Test for minimum compiler version, e.g. GCC 4.8 doesn't support std::regex.

### Approach

Test compiler version in cmake.

### Checklist


lochel:
I don't know what we should use as minimum version for clang.

OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/486//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#279

PR #294, opened 2018-08-22
missed one  dependent lib path

adrpo:
### Purpose

- fix the mac builds


OpenModelica-Hudson:

PR #302, opened 2018-08-23
proper disable ceres *and* dependencies

adrpo:
### Purpose

disable configuration of dependencies of ceres if is off


OpenModelica-Hudson:

PR #329, opened 2018-08-30
for cmake < 3.1 use CMAKE_CXX_FLAGS to set -std=c++11

adrpo:
### Purpose

- fix cmake setting of -std=c++11 for older cmake
- needed on debian jessie (cmake 3.0.2)


OpenModelica-Hudson:
All tests were successful. :smiley:

PR #354, opened 2018-09-19
Implement self referencing oms3_element_t type

lochel:
### Purpose

Extend the element information by its sub-elements.

### Approach

Implement `oms3_element_t` as self referencing struct.

### Type of Change

- Code refactoring (non-breaking change which improves maintainability)


lochel:
This should work, but I haven't tested it. @adeas31 can you test it easily using OMEdit?

I just realized that the API to delete a (sub-)system isn’t implemented. I will do that with a next PR.

PR #356, opened 2018-09-20
Save connectors to SSD

robbr48:
### Purpose

Save oms3::Connector to SSD

### Approach

- Call oms3::Connector::exportToSSD from oms3::System::exportToSSD
- Added oms3::Element::getConnectors function
- Added missing Elements tag when saving subsystems and components

### Type of Change

### Checklist


robbr48:
Sounds reasonable to me.

PR #357, opened 2018-09-20
Fix segfault in oms3::System::exportToSSD()

robbr48:
### Purpose

The connectors vector in oms3::System is null-terminated. Iterating over the elements causes a segmentation fault on last element.

### Approach

Check if each element exists before using it.

### Type of Change

### Checklist


lochel:
There is one more issue, see #354 (1d01e7b2ddd08410bbeb4303192518fd4309018c).

lochel:
Fixed as part of #354.

PR #363, opened 2018-09-20
Add lua binding to oms3_getSystemType

lochel:
### Purpose

Add lua binding to oms3_getSystemType.

### DIFF
--- a/src/OMSimulatorLib/Types.h
+++ b/src/OMSimulatorLib/Types.h
@@ -98,6 +98,7 @@ typedef enum {
 } oms_element_type_enu_t;
 
 typedef enum {
+  oms_system_none,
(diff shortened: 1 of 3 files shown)


OpenModelica-Hudson:

src/OMSimulatorLib/Types.h
```diff
@@ -98,6 +98,7 @@ typedef enum {
 } oms_element_type_enu_t;
 
 typedef enum {
+  oms_system_none,
```
adeas31: Good. I was thinking to add this.
Maybe we should also add `oms_element_none ` in `oms3_element_enu_t`.

lochel: Probably a good idea.

PR #376, opened 2018-09-25
Implement oms3_import

lochel:
### Purpose

Implement oms3_import

### Type of Change

### Open Questions and Pre-Merge TODOs

-  [x] Import ssd:SimulationInformation
-  [x] Import ssd:Connectors
-  [x] Import ssd:Connections

### DIFF
--- a/src/OMSimulatorLib/System.cpp
+++ b/src/OMSimulatorLib/System.cpp
@@ -259,6 +259,64 @@ oms_status_enu_t oms3::System::exportToSSD(pugi::xml_node& node) const
   return oms_status_ok;
 }
 
+oms_status_enu_t oms3::System::importFromSSD(const pugi::xml_node& node)
+{
+  for(pugi::xml_node_iterator it = node.begin(); it != node.end(); ++it)
+  {
+    std::string name = it->name();
+    if(name == oms2::ssd::ssd_simulation_information)
+    {
+      // TODO: import ssd_simulation_information
+    }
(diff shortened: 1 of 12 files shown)


lochel:
@adeas31 @robbr48 Do you think it is a good way of implementing the import feature? I am a bit uncertain if the code can be easily extended and maintained.

src/OMSimulatorLib/System.cpp
```diff
@@ -259,6 +259,64 @@ oms_status_enu_t oms3::System::exportToSSD(pugi::xml_node& node) const
...
+            return oms_status_error;
+
+          if (oms_status_ok != system->importFromSSD(*itElements))
+            return oms_status_error;
+        }
+        else
+          return logError("wrong xml schema detected: " + name);
+      }
+    }
+    else
```
robbr48: I think we should also load system annotations here.

lochel: Do we already export any system annotations?

robbr48: Not yet, but we will soon need them for bus connections.

OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/638//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#376
  - OpenModelica/OMSimulator-testsuite#84

OpenModelica-Hudson:

PR #431, opened 2018-10-24
Add simulation loop

lochel:
### Purpose

Add simulation loop

### Type of Change

### Checklist


src/OMSimulatorLib/SystemWC.h
```diff
@@ -63,7 +70,9 @@ namespace oms3
 
   private:
     std::string solverName = "oms-ma";
+    double time;
     double stepSize = 1e-1;
```
lochel: @robbr48 The old `communicationInterval` is the same then the new `stepSize`.

PR #432, opened 2018-10-24
Fix connectors of FMUs

lochel:
@adeas31 Not too nice, but should fix the issue.


OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/736//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#432

PR #460, opened 2018-11-01
Added TLMBusConnector connectors to C API

robbr48:
### Purpose

Make sub-connectors to TLM bus connectors available in the C API.

### Approach

Added two arrays of char*, one for name and one for TLM variable type. They are synchronized with the C++ map whenever adding new connectors.

### Type of Change

### Checklist


robbr48:
@adeas31 Is this what you need for OMEdit?

adeas31:
Looks good. Would be nice to free the memory in the destructor.

OpenModelica-Hudson:

PR #475, opened 2018-11-07
Export and import  element geometry to/from SSD

lochel:
### Related Issues

Reported by @adeas31 

### Purpose

Export and import  element geometry to/from SSD

### Type of Change

### DIFF
--- a/src/OMSimulatorLib/ComponentFMUCS.cpp
+++ b/src/OMSimulatorLib/ComponentFMUCS.cpp
@@ -270,6 +270,32 @@ oms3::Component* oms3::ComponentFMUCS::NewComponent(const pugi::xml_node& node, 
         component->connectors.push_back(oms3::Connector::NewConnector(*itConnectors));
       }
     }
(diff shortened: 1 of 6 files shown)


lochel:
@adeas31 I fixed also the delete function for system connectors, so that all associated connections get automatically removed.

adeas31:
I think we should not do that. OMEdit will not be able to add the connections back in the undo case then. If OMEdit deletes the connections then it can add them back on undo.

lochel:
Well, we do it also for systems and components.

lochel:
How do you undo the deletion of a system?

OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/828//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#475

OpenModelica-Hudson:
All tests were successful. :+1:

PR #501, opened 2018-11-14
Implement oms3_reset

lochel:
### Purpose

`oms3_reset` returns the model into the instantiated mode. That way the same model with different initialization can be run faster, because the fmus doesn’t need to be extracted and loaded.


OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/879//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#501

OpenModelica-Hudson:

Issue #496, opened 2018-11-13
OMEdit crashes when adding a CS-FMU to a strongly-coupled system


lochel:
OMEdit crashes when adding a CS-FMU to a strongly-coupled system.

I guess the same applies when adding a ME-FMU to a weakly-coupled system. The API should reject such operations.

PR #503, opened 2018-11-15
Fix several issues

lochel:
### Related Issues

#496

### Purpose

- [x] Check if FMU actually support specified FMI kind
- [x] Move result file implementation to namespace oms3
- [x] Fix bug for FMUs with very short name, e.g. A.fmu


OpenModelica-Hudson:
All tests were successful. :blush:

Issue #506, opened 2018-11-16
BuildSysPro.BaseClasses.HeatTransfer.Examples.ValidationLWRLinear.fmu fails with segmentation fault


lochel:
### Description

[BuildSysPro_BuildSysPro.BaseClasses.HeatTransfer.Examples.ValidationLWRLinear.sim](https://libraries.openmodelica.org/branches/master-fmi/BuildSysPro/files/BuildSysPro_BuildSysPro.BaseClasses.HeatTransfer.Examples.ValidationLWRLinear.sim)

```bash
Limited backtrace at point of segmentation fault
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12890)[0x7fe06b9e7890]
/var/lib/hudson/saved_omc/OMSimulator/install/linux/bin/OMSimulator(N_VMin_Serial+0x31)[0x56304498c92f]
/var/lib/hudson/saved_omc/OMSimulator/install/linux/bin/OMSimulator(N_VMin+0x24)[0x563044987a55]
/var/lib/hudson/saved_omc/OMSimulator/install/linux/bin/OMSimulator(CVodeSVtolerances+0xe6)[0x56304497ae20]
/var/lib/hudson/saved_omc/OMSimulator/install/linux/bin/OMSimulator(_ZN4oms38SystemSC10initializeEv+0x62b)[0x56304482cbcb]
/var/lib/hudson/saved_omc/OMSimulator/install/linux/bin/OMSimulator(_ZN4oms35Model10initializeEv+0x138)[0x5630448e4768]
/var/lib/hudson/saved_omc/OMSimulator/install/linux/bin/OMSimulator(oms3_initialize+0x51)[0x5630447dee11]
/var/lib/hudson/saved_omc/OMSimulator/install/linux/bin/OMSimulator(+0x123a3d)[0x563044790a3d]
/var/lib/hudson/saved_omc/OMSimulator/install/linux/bin/OMSimulator(main+0x9a1)[0x563044786bc1]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7)[0x7fe069db5b97]
/var/lib/hudson/saved_omc/OMSimulator/install/linux/bin/OMSimulator(_start+0x2a)[0x56304478a90a]
```

### Steps to reproduce the behavior

```bash
[...]

PR #508, opened 2018-11-16
Fix SC systems without continuous states

lochel:
### Related Issues

#506 

### Purpose

Fix SC systems without continuous states


OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/885//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#508

Issue #528, opened 2018-12-01
setLoggingLeve() 


meek1:
Seems like setLoggingLevel() is not yet implemented for oms3? Needed for debugging.

PR #534, opened 2018-12-12
Implement oms3_setLoggingLevel

lochel:
### Related Issues

#528

### DIFF
--- a/src/OMSimulatorLib/Logging.cpp
+++ b/src/OMSimulatorLib/Logging.cpp
@@ -57,7 +57,7 @@ Log::Log() : filename(""), cb(NULL)
   numWarnings = 0;
   numErrors = 0;
   numMessages = 0;
-  logLevel = false;
+  logLevel = 0;
 }
 
 Log::~Log()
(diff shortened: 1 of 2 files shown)


OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/920//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#534

OpenModelica-Hudson:

robbr48:
This should also be propagated to OMTLMSimulator.

lochel:
What does it mean? Is TLM simulator using a different logging system?

robbr48:
Yes, it writes to TLMlogfile.log. 

For oms2 we had a special `oms2_setTLMLoggingLevel` API function, but perhaps we can translate the OMSimulator log level to the `TLMLogLevel` used in OMTLMSimulator.

Available log levels are: Disabled, Fatal, Warning, Info, Debug

lochel:
I think we should use the OMSimulator logging system for the TLM simulator. Maybe we can register a callback function to handle the logging, similar to the FMIL. That way SKF would not depend on it.

PR #561, opened 2019-01-14
link with -latomic on i386

adrpo:
### Purpose

- fix builds on Linux i386


OpenModelica-Hudson:

PR #578, opened 2019-01-21
Read version string from file if git is not available

lochel:
### Related Issues

[ticket:5246](https://trac.openmodelica.org/OpenModelica/ticket/5246)

### Purpose

Read version string from file if git is not available


OpenModelica-Hudson:
All tests were successful. :smile:

Issue #559, opened 2019-01-14
Add Linux builds


meek1:
Please add pre-compiled binaries for "generic Linux usage" on github (as already done for Windows). Should include both native FMI and TLM-functionality, and work for e.g. CentOS6 and 7. SysIdent may be omitted if problematic.

meek1:
@sjoelund, @lochel Any progress on this one?

sjoelund:
I have made some progress. Still trying to figure out how to not link some dynamic versions of system libraries in cmake. However, @lochel would need to remove the boost dependencies or add them to 3rdParty and compile static versions using -fPIC if we want to use the Python interface. It seems all Linux distributions compile their static Boost libraries without -fPIC, which only works on 32-bit OS.

sjoelund:
#586 adds Linux builds to [our Jenkins](https://test.openmodelica.org/jenkins/job/OMSimulator/job/master/lastCompletedBuild/artifact/) (not last successful build because @lochel doesn't use our Jenkins, but our Hudson and as such the tests are usually slightly broken; it will take a while before this is updated, so I will attach a static build to this comment: [OMSimulator-linux-amd64-v2.1.0-dev-10-g5c19e85.tar.gz](https://github.com/OpenModelica/OMSimulator/files/2787090/OMSimulator-linux-amd64-v2.1.0-dev-10-g5c19e85.tar.gz))

Note that it is built using MUSL C and not GNU C since this is much easier with static linking. However, the TLM Simulator does not seem to support it. I did not try OMSysIdent. I tested the binary on CentOS6 (among others):
```bash
$ sudo docker run -it -w $HOME/Downloads -v $HOME/Downloads:$HOME/Downloads centos:6 ./bin/OMSimulator -v
OMSimulator v2.1.0-dev-10-g5c19e85-linux-notlm
```

The alpine:edge OS had boost with -fPIC so I could compile it with Boost and not rely on C++17 or rewriting existing code.

lochel:
We are now using Jenkins and Windows, Linux, and Mac builds are available here: https://test.openmodelica.org/jenkins/job/OMSimulator/job/master/lastSuccessfulBuild/artifact/

This is now also linked from the README.

lochel:
Can this be closed?

sjoelund:
[...]

PR #586, opened 2019-01-23
Make a static build using Alpine Linux

sjoelund:
Alpine Linux uses MUSL libc, which is much easier to statically link
than GNU libc. libOMSimulator.so is not compiled due to Boost .a-files
not being compiled with -fPIC.

### Related Issues

#559

### DIFF
--- a/.CI/alpine/Dockerfile
+++ b/.CI/alpine/Dockerfile
@@ -1,0 +1,4 @@
+# We need gcc 8 for filesystem support
+FROM alpine:edge
+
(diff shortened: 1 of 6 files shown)


OpenModelica-Hudson:
The following pull requests are now being [tested](https://test.openmodelica.org/hudson/job/OMSimulator-PR/1028//console). It may take a while, depending on the workload.
  - OpenModelica/OMSimulator#586

OpenModelica-Hudson:
All tests were successful. :smile:

Issue #168, opened 2018-05-15
Communication step size adaption


lochel:

meek1:
Any news from thesis work? Maybe this issue can be moved to in progress and broken downd to sub tasks.

PR #602, opened 2019-01-30
Adaptive step-size solver for Co-simulation

frape315:
### Related Issues

* New feature "Adaptive step-size solver"
* #168

### Purpose

This is the branch on which the adaptive step-size solver has been designed on. 

### Approach

- Added an extra flag DoubleStep. 
- New Solver method 'oms_solver_wc_mav' which is an adaptive solver implemented in SystemWC.
- To Enable this adaptive solver oms_setSolver(<model/System>, oms_solver_wc_mav)
- It adds this solver as an extra option, by default fixed solver will be used still. (if you wish to force fixed solver use  oms_setSolver(<model/System>, oms_solver_wc_ma)
- Best results are used when adding flag --setInputDerivatives=true
- Two seperate algorithms, by default it will attempt to use single step algorithm --doubleStep=true makes it use double-step if possible (only possible if all FMUs can get/Set state).

### Type of Change

- New feature
- Breaking change (fix or feature that could cause existing functionality to not work as expected)

### Checklist

### DIFF
--- a/src/OMSimulatorLib/Flags.cpp
+++ b/src/OMSimulatorLib/Flags.cpp
@@ -70,6 +70,7 @@ void oms::Flags::setDefaults()
   realTime = false;
   resultFile = "<default>";
   solver = oms_solver_sc_cvode;
+  solverStats = false;
   startTime = 0.0;
   stopTime = 1.0;
   stripRoot = false;
@@ -278,6 +279,12 @@ oms_status_enu_t oms::Flags::Solver(const std::string& value)
   else
     return oms_status_error;
 
+  return oms_status_ok;
+}
+
+oms_status_enu_t oms::Flags::SolverStats(const std::string& value)
+{
+  GetInstance().solverStats = (value == "true");
   return oms_status_ok;
 }
 
--- a/src/OMSimulatorLib/Flags.cpp
+++ b/src/OMSimulatorLib/Flags.cpp
@@ -62,6 +62,7 @@ oms::Flags::~Flags()
 void oms::Flags::setDefaults()
 {
   defaultModeIsCS = false;
+  doubleStep = false;
   ignoreInitialUnknowns = false;
   inputDerivatives = false;
   intervals = 100;
@@ -335,8 +336,15 @@ oms_status_enu_t oms::Flags::WorkingDir(const std::string& value)
   return oms_status_ok;
 }
 
+
 oms_status_enu_t oms::Flags::WallTime(const std::string& value)
 {
   GetInstance().wallTime = (value == "true");
+  return oms_status_ok;
+}
+
+oms_status_enu_t oms::Flags::DoubleStep(const std::string& value)
+{
+  GetInstance().doubleStep = (value == "true");
   return oms_status_ok;
 }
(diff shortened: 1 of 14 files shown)


src/OMSimulatorLib/Flags.cpp
```diff
@@ -66,7 +66,11 @@ void oms::Flags::setDefaults()
   progressBar = false;
   inputDerivatives = false;
   defaultModeIsCS = false;
+<<<<<<< HEAD
   wallTime = false;
+=======
+>>>>>>> c512e8c56d33d4b11b8cd3498ef49bb77608c0f4
```
lochel: Seems like something went wrong while merging conflicts.

lochel:
@frape315 Please rebase the branch.

src/OMSimulatorLib/Model.h
```diff
@@ -129,6 +130,15 @@ namespace oms
     int bufferSize = 10;
     std::string resultFilename;             ///< default <name>_res.mat
     Clock clock;
+    unsigned int clock_id;
```
lochel: The model shouldn't have a `$wallTime`-clock.

src/OMSimulatorLib/Flags.h
```diff
@@ -129,7 +131,8 @@ namespace oms
       {"--tolerance", "", "Specifies the relative tolerance", re_double, Flags::Tolerance, false},
       {"--version", "-v", "Displays version information", re_void, Flags::Version, false},
       {"--wallTime", "", "Add wall time information for to the result file", re_bool, Flags::WallTime, false},
-      {"--workingDir", "", "Specifies the working directory", re_default, Flags::WorkingDir, false}
+      {"--workingDir", "", "Specifies the working directory", re_default, Flags::WorkingDir, false},
+      {"--doubleStep", "", "Changes the algorithm inside the WC sytems to double-step when using adaptive step-size", re_bool, Flags::DoubleStep, false}
```
lochel: The double-step-master should be implemented as another solver method, instead of using the flag.

src/OMSimulatorLib/Model.cpp
```diff
@@ -587,6 +595,25 @@ oms_status_enu_t oms::Model::emit(double time, bool force)
...
+    resultFile->updateSignal(h_id,stepS);
+    SignalValue_t rollB;
+    rollB.intValue = rollBackIt;
+    resultFile->updateSignal(roll_iter_id,rollB);
+    SignalValue_t maxErr;
+    maxErr.realValue = maxError;
+    resultFile->updateSignal(max_error_id,maxErr);
+    SignalValue_t errNorm;
+    errNorm.realValue = normError;
+    resultFile->updateSignal(error_id,errNorm);
```
lochel: This additional information shouldn't be available by default. Please add a flag to enable it.

lochel: … and they need to be moved to oms::SystemWC

lochel:
@adeas31 This introduces another master algorithm (`oms_solver_wc_mav2`) which need to be added to OMEdit.

adeas31:
OpenModelica/OMEdit#218 takes care of it.

lochel:
The algorithm doesn't work as expected. Probably I have introduced some bugs. @frape315 can you have a look?

PR #608, opened 2019-02-07
New flag --stripRoot

lochel:
### Purpose

Add a new flag to supress the `<mode>.<root-system>.` prefix for all signals in the result file.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,7 @@ pipeline {
   }
   options {
     newContainerPerStage()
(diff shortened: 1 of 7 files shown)


lochel:
@robbr48 @sjoelund The tlm tests are still failing randomly. 😒

sjoelund:
Yes, they seem to hang or crash a lot especially on Ubuntu 18.04. We would probably need a retry mode for TLM tests unless it is resolved by @robbr48 soon.

PR #613, opened 2019-02-08
Fix bug in TLMBusConnector::getActualBus

robbr48:
### Purpose

Fix crash in some TLM test models due to non-connected TLM buses which didn't identify themselves as their "actual bus".

### Approach

Return self instead of nullptr if there are no connections in the system that contains the TLM bus.

### Type of Change

### Checklist

### DIFF
--- a/src/OMSimulatorLib/TLM/TLMBusConnector.cpp
+++ b/src/OMSimulatorLib/TLM/TLMBusConnector.cpp
@@ -342,7 +342,7 @@ oms::TLMBusConnector* oms::TLMBusConnector::getActualBus(ComRef cref, System *sy
       subcref = connections[i]->getSignalA();
     }
     else {
-      return this;
+      continue;
     }
 
     TLMBusConnector* subBus = system->getTLMBusConnector(subcref);
@@ -359,7 +359,7 @@ oms::TLMBusConnector* oms::TLMBusConnector::getActualBus(ComRef cref, System *sy
     }
   }
 
-  return nullptr; //Should never happen
+  return this;
 }
 
 std::vector<std::string> oms::TLMBusConnector::getVariableTypes(oms_tlm_domain_t domain, int dimensions, oms_tlm_interpolation_t interpolation)


lochel:
There are more problems with TLM: https://test.openmodelica.org/jenkins/blue/organizations/jenkins/OMSimulator/detail/PR-613/1/tests

sjoelund:
That particular problem could be fixed by in `omtlm_checkPortAvailability` saying that if you can't figure out what address family you're using, assume AF_INET (IPv4 or IPv6).
Actually, the entire code assumes AF_INET as it's the only supported option so it could just... I dunno, skip the tests and just put AF_INET in there? :)

lochel:
> That particular problem could be fixed by in omtlm_checkPortAvailability saying that if you can't figure out what address family you're using, assume AF_INET (IPv4 or IPv6).
> Actually, the entire code assumes AF_INET as it's the only supported option so it could just... I dunno, skip the tests and just put AF_INET in there? :)

This is already addressed by some other PR, right?

robbr48:
@lochel I believe this can be merged now.

lochel:
It seems the Jenkins job hangs ...

lochel:
I've restarted it now.

robbr48:
Now there is some weird socket error. `gethostname`returns the name "97b517ed1550", which is not found by `gethostbyname`...

sjoelund:
Kubernetes instances don't have a hostname that resolves to anything :)

I fixed that pattern in one place. It seems it used elsewhere as well... It was just used to find out that the localhost had an IP address...

robbr48:
I see, it was the same problem on a different place. I resolved it by OpenModelica/OMTLMSimulator#113. The tests should hopefully work now.

lochel:
It still doesn't seem to work.

robbr48:
I think I found the problem. We use `gethostname` followed by `gethostbyname` for determining the local ip of the TLM manager. This is only used for external models, since FMUs are run locally.  Hence, "tlmexternal.lua" fails since `gethostname` does not work on the test server. I will see if I can change the code somehow.

robbr48:
Solved in OpenModelica/OMTLMSimulator#116 (uses specified address, so there is no need to call `gethostname` anymore)

robbr48:
My bad. The fix had no effect due to the wrong operator in an if statement condition. I changed it in OpenModelica/OMTLMSimulator#117.

Issue #562, opened 2019-01-15
Implement oms2 features for oms3


lochel:
The old oms2 implementation is obsolete and will be removed for the v2.0 release. However, not everything is implemented in the new API yet. It is fine to do some things a bit later.

I open this issue so that we do not forget about some features.

- [x] soft real-time functionality (@bernhard-thiele)
- [x] parallel master algorithm (@bernhard-thiele)
- [x] adaptive step size control developed by @farkasrebus
- [x] OMSysIdent

lochel:
For the time being, all oms2 development will stay available from the oms2 branch.

PR #614, opened 2019-02-08
Re-add adaptive step-size control

lochel:
### Related Issues

#562

### Purpose

Re-add adaptive step-size control developed by @farkasrebus.


lochel:
@farkasrebus Everything compiles now with the new implementation. However, the simulation produces different results. Can you have a look?

farkasrebus:
Line 18 of testsuite/assc/ASSCExample.lua should be changed to
`oms_addConnection("Example.root.Class1.max", "Example.root.Class2.max")`

But that doesn't solve the problem, so I'll keep investigating.

farkasrebus:
I think `oms::SystemWC::stepUntilASSC()` is never called.  If I understand the code correctly `oms_simulate()` calls `oms::Model::simulate()` that calls `oms::SystemWC::stepUntil()` directly, which performs fixed step size simulation with the maximal step size.

lochel:
@farkasrebus Thanks for the hints. I fixed everything you have mentioned, but the results are still different. Can you please have another look?

farkasrebus:
The step size controller operates correctly, but the FMUs don't behave as expected. Replacing the FMUs with new ones generated manually with OMEDit solved the problem for me.

More specifically the signal `signalwiththreshold` of Class2 is only supposed to increase up to `max`. In the simulated scenarario there is a _communication time point_ at 2 s, where `signalwiththreshold=max`, so therefore the value of the signal should not change until `max` is modified  but in the simulation it keeps increasing for one more step. Same thing happens again when the maximum is reached for the second time.

lochel:
@farkasrebus  I just tried it again, but I get the same wrong behaviour with the new FMUs. Can you please push your local changes to this branch?

lochel:
@farkasrebus Thanks!

PR #618, opened 2019-02-11
Enable tlm tests again

lochel:
This reverts partially aca92db88f8d6a8f8cc4950a18792d030548a7ac.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,7 @@ pipeline {
   parameters {
     booleanParam(name: 'MSVC64', defaultValue: false, description: 'Build with MSVC64 (often hangs)')
     booleanParam(name: 'MINGW32', defaultValue: false, description: 'Build with MINGW32 (does not link boost)')
-    string(name: 'RUNTESTS_FLAG', defaultValue: '-fast', description: 'runtests.pl flag')
+    string(name: 'RUNTESTS_FLAG', defaultValue: '', description: 'runtests.pl flag')
   }
   stages {
     stage('build') {


sjoelund:
Seems you got a machine with the hostname issue we talked about earlier

lochel:
Is that fixed now?

sjoelund:
Should be

sjoelund:
Wouldn't it be simple to keep the flag though? But keep the default value of the parameter as empty?

lochel:
@robbr48 Some tlm tests are still failing. Can you please have a look?

robbr48:
@sjoelund @lochel 
The segmentation fault occurs on the following line (TLMClientComm.cc, 210):
    `memcpy((char *)&sa.sin_addr, hp->h_addr, hp->h_length);`
Where `hp` is obtained from
    `hp = gethostbyname(callname.c_str());`
Can it be related to the other problem with `gethostbyname()`?

sjoelund:
What is `callname.c_str()` in this case? I first suspected this could be that you do `gethostbyname()` and get an INET6 address, but it should only return INET4 addresses...
The only thing I could think of is if `hp->h_addr` is NULL, which is not checked.

robbr48:
`callname.c_str()` is printed to the logfile and looks ok. In this case it is "127.0.0.1". I also printed `hp->h_addr`, and it is not NULL. I'm thinking that it could have something to do with thread safety, since multiple threads are calling `gethostbyname` in parallel. Perhaps it is not thread safe.

sjoelund:
Yes, `gethostbyname` is marked thread-unsafe in the man-pages (`MT-Unsafe race:hostbyname env locale`)
`gethostbyname_r` exists but is a bit trickier to use it seems (and is a GNU extension).

sjoelund:
If you know the string is just an IPv4 address and never a hostname, you could also use `inet_addr()` instead.
You could also
* Create a global lock for the host lookup
* Lookup all host names before creating the threads

robbr48:
Ok, I solved it by adding mutex locks, since it should be possible to use hostnames (#623)

lochel:
The tlm tests are unfortunately still unstable:
```bash
+ tlm1d.lua ... execution failed

==== Log /tmp/oms-rtest-unknown/tlmtmp_5773/log-tlm1d.lua
Starting TLM simulation.
Monitoring thread finished.
Fatal error: Failed to send message header. Aborting.
info: Logging information has been saved to "tlm1d.log"
```

robbr48:
@lochel Yes, that's a different error. It should be solved now by OpenModelica/OMTLMSimulator#112

lochel:
The test tlm1d.lua is failing with a segmentation fault.

```bash
+ tlm1d.lua ... execution failed

==== Log /tmp/oms-rtest-unknown/tlmtmp_9083/log-tlm1d.lua
Segmentation fault (core dumped)
```

PR #631, opened 2019-02-14
New API to load a snapshot

lochel:
### Related Issues

The undo functionality in OMEdit depends on loading previous model states.

### Purpose

New API to load a snapshot.

### Approach

A snapshot is the ssd representation of a model. The new API can be used to reset or update an existing model.

### Type of Change


lochel:
@adeas31 I expect it now to work now. The new API can be used to restore a model state, e.g. undo operations, and for text editing of the ssd representation in OMEdit. However, the text representation will be reformatted.

PR #632, opened 2019-02-14
Fix nested namespaces

lochel:
### Related Issues

Fix builds using OMDev, which doesn't support C++17.


lochel:
TLM tests failed again.

PR #607, opened 2019-02-07
Run wine of the MinGW tests

sjoelund:
### Purpose

Run wine for the Windows tests

PR #635, opened 2019-02-14
Disable TLM 3D testmodel

robbr48:
### Purpose

TLM 3D test model seem to run forever on Jenkins. Disable it for now.

### Approach

Remove it from Makefile (add it to FAILINGTESTS)

### DIFF
--- a/testsuite/tlm/Makefile
+++ b/testsuite/tlm/Makefile
@@ -5,7 +5,6 @@ externalmodels.lua \
 externalmodels.py \
 tlm1d.lua \
 tlm1dfg.lua \
-tlm3d.lua \
 tlmbuses.lua \
 tlmbuses.py \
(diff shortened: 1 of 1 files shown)


robbr48:
Seems to work. @lochel @sjoelund can we merge this?

sjoelund:
I suppose we could. I'd like a fix for the issue seen in #607 though (gethostbyname is called in manager thread still, right?)

robbr48:
I removed all calls to `gethostbyname` in OpenModelica/OMTLMSimulator#113, OpenModelica/OMTLMSimulator#114 and OpenModelica/OMTLMSimulator#115. It is still used in one place for printing the server name, but that is not used by OMSimulator and never executed in parallel. 

However, it is still used in multiple places for WIN32. We may need to look over this later.

PR #636, opened 2019-02-14
Update OMTLMSimulator

sjoelund:
Avoid crashes during startup.


lochel:
This is pushed as part of #637.

PR #639, opened 2019-02-19
Reduce port numbers for two TLM test models

robbr48:
### Purpose

TLM 3D test model seem to run forever on Jenkins. It is using the highest port number.

### Approach

Test and reduce initial port number.

### DIFF
--- a/testsuite/tlm/tlm3d.lua
+++ b/testsuite/tlm/tlm3d.lua
@@ -170,7 +170,7 @@ oms_addConnectorToTLMBus("model.tlm3d.wc2.P","model.tlm3d.wc2.t3", "effort6");
 
 oms_addTLMConnection("model.tlm3d.wc1.P","model.tlm3d.wc2.P", 0.00001, 0.9, 10000, 100);
 
-oms_setTLMSocketData("model.tlm3d","127.0.1.1",13511,13611)
+oms_setTLMSocketData("model.tlm3d","127.0.1.1",11511,12511)
 
 oms_setStartTime("model",0);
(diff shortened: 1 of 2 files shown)


robbr48:
Seems to work. @sjoelund can it be that ports 13511/13611 were not available on the test server?

sjoelund:
The ASAN tests I believe run directly on the machine (actually docker in privileged mode) so if we allow concurrent tests the same ports might be used I suppose. All other tests run on its separate network.

robbr48:
That's the reason why all TLM test models have different port numbers. They also increment the number automatically until they find an available port. Therefore I added an offset of 100 between each model. 

In this case though, it seems like the 3D TLM model failed because it was using a too high number, which I cannot really explain. Can we restart the tests to confirm this?

lochel:
I've restarted the tests.

lochel:
All tests were successful. Anyway, I've restarted the tests again.

robbr48:
Looks like it is stuck again :( I guess there is some other problem with the 3D model then. Is there any possibility to view the log files?

robbr48:
Running the tlm3d.lua model causes OMSimulator to use more than 1GB of memory. Can this be a problem on Jenkins?

sjoelund:
Not likely. The docker image has 6GB pre-allocated RAM for each instance running (for Kubernetes; the other machines have more RAM than that available; usually at least 16GB)...

robbr48:
Then I guess we should simply disable it for now (#635). I cannot figure out how to solve the problem.

Issue #157, opened 2018-05-09
Oscillations in FMI co-simulation


meek1:
Oacillating behavior for native FMI-based (not TLM) co-simulation of aricraft use-cases. Might be related to issue #18 "Coupling stability".

meek1:
@lochel, aslo oms3 shows oscillations for the aircraft demo. However not exactly the same for 32- and 64-bit.

![image](https://user-images.githubusercontent.com/31110788/48678081-d8c50f80-eb7e-11e8-91bb-15243d93ac4a.png)

lochel:
It seems that this issues can finally be closed. The following three signals were produced without input extrapolation (blue, tol=1e-3), with input extrapolation (red, tol=1e-3), with extrapolation and tighter tolerances (green, tol=1e-6).

![image](https://user-images.githubusercontent.com/8457302/55630171-59467e80-57b5-11e9-8f60-a1105603e556.png)

PR #651, opened 2019-04-05
Handle derivatives that are NAN or INF

lochel:
### Related Issues

#157 

### Purpose

Sometimes FMUs return NAN or INF as the output derivative.

### Approach

Override those derivatives with zero and print a warning.


sjoelund:
Wouldn't it be better to take a shorter step instead (and then possibly fail after some tries)? It gives great improvements for the library coverage, but I fear the results would be wrong:
https://libraries.openmodelica.org/branches/history/master-fmi/2019-04-02%2002:56:54..2019-04-05%2019:49:39.html

lochel:
This shouldn't have any impact on the library coverage, because it tests only independent FMUs.

I noticed the NAN/INF problem only right after initialization (before the first step) of some FMUs using numeric derivatives. Therefore, adjusting the step size won’t help.

PR #673, opened 2019-05-07
Update OMTLMSimulator submodule

robbr48:
### Related PR

OpenModelica/OMTLMSimulator#118

### Purpose

Submodule update


lochel:
I restarted the Jenkins job to test the updated submodule.

lochel:
The hash is wrong. It points to the commit from the tlm PR, and not to the one that is actually in the tlm master.

PR #682, opened 2019-05-16
 Add test case for adaptive step size master

lochel:
### Purpose

This is supposed to test the adaptive step size master, but OpenModelica FMUs doesn't yet support this functionality.


meek1:
@lochel, just curious, is this for the adaptive step size master based on work by @frape315?

lochel:
Yes, but unfortunately OpenModelica doesn't support get/set FMU state and we don't have licence free Dymola FMUs. Therefore, this test is a placeholder until OpenModelica provides the require FMI functionality.

PR #686, opened 2019-05-27
Catch more filesystem exceptions

lochel:
### Related Issues

UTF-8 characters in path specifiers cause exceptions in the file system API.

### Purpose

Catch the exception instead of crashing the application - and therewith also OMEdit.

### Further steps

* Support utf8 path specifiers

### DIFF
--- a/src/OMSimulatorLib/Scope.cpp
+++ b/src/OMSimulatorLib/Scope.cpp
@@ -220,61 +220,83 @@ oms_status_enu_t oms::Scope::importModel(const std::string& filename, char** _cr
 
 oms_status_enu_t oms::Scope::setTempDirectory(const std::string& newTempDir)
 {
-  if (!filesystem::is_directory(newTempDir))
-  {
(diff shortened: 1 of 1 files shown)


lochel:
I can run OMSimulator from an utf8 path using mingw on Windows. It creates files and directories in this path without crashing.

```bash
lenna@LACO-BOOK MINGW64 /c/Users/lenna/Desktop/白树
$ OMSimulator.exe --logFile=. --workingDir=. --tempDir=./tmp/
info:    Set temp directory to    "C:/Users\lenna\Desktop\??"
info:    Set working directory to "C:/Users\lenna\Desktop\??"
info:    Set working directory to "C:/Users\lenna\Desktop\??"
info:    New temp directory has been created: "./tmp/"
info:    Set temp directory to    "C:/Users\lenna\Desktop\??\tmp"
```

adeas31:
The try catch code blocks looks fine but is the example output you posted is the expected one? What happens if you call `oms_newModel`? Where is the model directory created?

PR #690, opened 2019-06-20
Fix OMSysIdent C and C++ example

bernhard-thiele:
### Purpose

The two examples `test_HelloWorld_cs_Fit.c` and `test_Lin2DimODE_cs_Fit.cpp` were broken since FMUs were moved into another folder and another scheme for accessing variables was introduced.

### Approach

Adapting paths and names


bernhard-thiele:
I also fixed the other non-working C++ example, modified the descripiton, rebased the work on the updated master branch and force-pushed everything.

Issue #693, opened 2019-06-21
Integrate OMSysIdent C/C++ examples into CI tests


bernhard-thiele:
**Is your feature request related to a problem? Please describe.**
Currently there are two OMSysIdent test examples (`test_HelloWorld_cs_Fit.c` and `test_Lin2DimODE_cs_Fit.cpp`) which can be executed using `ctest`, but those tests are not within the CI test suite.  A problem is that the test suite only has direct support for Lua or Python scripts.

**Describe the solution you'd like**
Wrap the `ctest` execution within a Python script in way so that the existing CI test suite can be used without modification (this is my current solution).

**Describe alternatives you've considered**
It should also be possible to adapt the CI test suite and have a more direct support for C/C++ tests and/or `ctest` integration. This would probably be nicer, but it is not obvious how to do it best and it would require more work.

PR #695, opened 2019-06-21
Clean up after OMSysIdent test runs

bernhard-thiele:
### Related Issues

#693

### Purpose

Clean up the generated temporary files after running the OMSysIdent tests.

### Approach

Using the `teardown_command` functionality for removing respective files after a test run.


testsuite/OMSysIdent/runCTests.py
```diff
@@ -1,8 +1,12 @@
 ## status: correct
-## teardown_command: rm test_HelloWorld_cs_Fit.log test_HelloWorld_cs_Fit_res.mat test_Lin2DimODE_cs_Fit.log test_Lin2DimODE_cs_Fit_res.mat
+## teardown_command: rm -rf test_HelloWorld_cs_Fit/ test_HelloWorld_cs_Fit.log test_HelloWorld_cs_Fit_res.mat test_Lin2DimODE_cs_Fit/ test_Lin2DimODE_cs_Fit.log test_Lin2DimODE_cs_Fit_res.mat
```
lochel: This test should also state explicitly that it is only designed to run on linux:
```
## linux: yes
## mingw: no
## win: no
## mac: no
```

lochel:
Thanks 👍

PR #700, opened 2019-07-05
Expose external model info through API

robbr48:
### Purpose

Provide information about external models from C API.

### Approach

A struct with information, similar to how to FMU info is handled. Only path and start script is currently included, but it will be extended later.


src/OMSimulatorLib/DummyTLM/TLMInterface.h
```diff
@@ -31,7 +31,7 @@
 
 #ifndef _DUMMY_TLM_INTERFACE_H_
 #define _DUMMY_TLM_INTERFACE_H_
-#define NO_TLM 1
+//#define NO_TLM 1
```
lochel: Why is it commented out?

robbr48: My bad. I only commented it out to temporary force enable TLM in my build environment.

robbr48:
@lochel I removed the submodule update as well.

lochel:
> @lochel I removed the submodule update as well.

Good, since it otherwise Jenkins isn't going to test the changes,

src/OMSimulatorLib/Component.h
```diff
@@ -69,6 +70,7 @@ namespace oms
     const std::string& getPath() const {return path;}
     oms_component_enu_t getType() const {return type;}
     virtual const FMUInfo* getFMUInfo() const {return NULL;}
+    virtual const ExternalModelInfo* getExternalModelInfo() const {return NULL;}
```
lochel: Not sure if we should really have `getFMUInfo` and  `getExternalModelInfo` in class Component or not... 🤔

robbr48: I guess there is no need, since each of these functions are only used in a single derived class. I can fix it when I'm back from my summer vacation.

robbr48: I removed `getExternalModelInfo` from `Component` class. I'm not sure how to solve it for `getFMUInfo`, since it is used in two subclasses. It doesn't really belong to this PR anyway, so I leave it for now.

robbr48:
@lochel I believe this can be merged now. I am still not authorized.

src/OMSimulatorLib/ExternalModelInfo.cpp
```diff
@@ -0,0 +1,48 @@
...
+ * FOR A PARTICULAR PURPOSE, EXCEPT AS EXPRESSLY SET FORTH
+ * IN THE BY RECIPIENT SELECTED SUBSIDIARY LICENSE CONDITIONS OF OSMC-PL.
+ *
+ * See the full OSMC Public License conditions for more details.
+ *
+ */
+
+#include "ExternalModelInfo.h"
+
+#include "Logging.h"
```
adeas31: I guess this one is not required.

src/OMSimulatorLib/ExternalModelInfo.h
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+
+#ifndef _OMS_EXTERNAL_MODEL_INFO_H_
+#define _OMS_EXTERNAL_MODEL_INFO_H_
+
+#include "Types.h"
+
+#include <string>
+#include <fmilib.h>
```
adeas31: `fmilib.h` is probably also not required here.

src/OMSimulatorLib/ExternalModelInfo.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace oms
+{
+  /**
+   * \brief FMU info
+   */
```
adeas31: Please update the comment to `External model info`.

robbr48: Thanks for the input, it is corrected now!

robbr48:
@adeas31 @lochel Can this be merged now?

adeas31:
@lochel is on vacations. To me this looks good so I am going to merge it in.

Issue #202, opened 2018-05-30
Fault Injection Block


meek1:
To facilitate what-if simulations, the master should include a Fault Injection Block (FIB). The FIB is a master functionality that can be inserted to modify inputs of FMUs. An example of a typical FIB functionality is found in Section 3.5 of https://www.diva-portal.org/smash/get/diva2:474629/FULLTEXT01.pdf

The FIB could be an optional input argument to oms2_addConnection, e.g.:
oms2_addConnection("FMICompositeModel","FMU_A:u1","FMU_B:y1", [faultType faultValue]), with faultType as Integer and faultValue as Real. Then input FMU_A.u1 is modified according to faultType and faultValue.

lochel:
If I understood it correctly, this functionally could be easily realised using an FMU as fault injection block. It basically would have one input and one output and put in the middle of the connection(s) to investigate. It could support all the different fault types described in the document.

lochel:
Should we assign a FIB to an output, input, or connection?

**Proposal**
One option would be to directly assign the fault injection to an output. The original output, e.g. `model.root.unit.y`, could then be redirected to `model.root.unit.y.$original`, and `model.root.unit.y` would be calculated based on its original value and the fault type.

API: `oms_faultInjection(<output name>, <fault type>, <fault value>)`.

meek1:
Yes I think it shoul be on the output, seems to be in line with the reference mentioned above. A typical example is to simulate sensor failure.

meek1:
I think current proposal above (on output) is good.

lochel:
I will consider the following fault types:
```
typedef enum {
  oms_fault_type_none,      ///< y := y.$original
  oms_fault_type_bias,      ///< y := y.$original + faultValue
  oms_fault_type_gain,      ///< y := y.$original * faultValue
  oms_fault_type_const      ///< y := faultValue
} oms_fault_type_enu_t;
```

lochel:
It is now available for testing. Documentation and Python bindings will be added soon.

lochel:
This is how to create an fib:
```
--oms_faultInjection("PI_Controller.co_sim.I.y", oms_fault_type_bias, 0.2)
oms_faultInjection("PI_Controller.co_sim.I.y", oms_fault_type_gain, 10.0)
```

PR #703, opened 2019-07-17
Fault Injection Block

lochel:
### Related Issues

#202 

### Purpose

Support fault injection block.


lochel:
I'm back from vacation and will continue to work on this now.

lochel:
Fault injection block are now supported for CS FMUs. However, some more testing is needed before it can be merged.

PR #704, opened 2019-09-02
API for fetching external model interface data

robbr48:
### Related PRs
OpenModelica/OMTLMSimulator#119

### Purpose

External models are not loaded as model files. Thus, interfaces needs to be fetched through socket communication. This adds an API function for this, which returns a list of names, domains and dimensions for all interfaces in an external model.

### Approach

The existing API for this in OMTLMSimulator i used. The external tool is instructed by sockets to write an XML file, which OMSimulator in turn reads.

### DIFF
--- a/src/OMSimulatorLib/TLM/SystemTLM.cpp
+++ b/src/OMSimulatorLib/TLM/SystemTLM.cpp
@@ -142,8 +142,11 @@ oms_status_enu_t oms::SystemTLM::initialize()
   omtlm_setMonitorPort(model, actualMonitorPort);
 
   for(const auto& subsystem : this->getSubSystems()) {
+    if(!singleModel.empty() && singleModel != subsystem.second->getCref().c_str()) {
+        continue;
+    }
     omtlm_addSubModel(model, subsystem.second->getCref().c_str(),"","none");
-    for(int i=0; subsystem.second->getTLMBusConnectors()[i]; ++i) {
+    for(int i=0; subsystem.second->getTLMBusConnectors()[i] && (singleModel==""); ++i) {
       TLMBusConnector *tlmbus = subsystem.second->getTLMBusConnectors()[i];
 
       std::string causality = "input";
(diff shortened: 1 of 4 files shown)


robbr48:
@lochel It seems like Jenkins refuses to build the PR because of the submodule update. It says I should set the SUBMODULE_UPDATE option, but I still don't have any Jenkins credentials.

lochel:
I will trigger the job for you.

lochel:
@robbr48 please contact @sjoelund to get credentials.

sjoelund:
I have added credentials to account @robbr48 on Jenkins

robbr48:
@lochel @sjoelund Thanks!

robbr48:
@lochel I am also not allowed to merge, or even assign reviewers :) Can you have a look at this and/or merge it?

sjoelund:
Developers only had read access before (since Hudson merged for us). I changed this to write access.

src/OMSimulatorLib/TLM/SystemTLM.h
```diff
@@ -112,6 +113,7 @@ namespace oms
     double logStep = 1e-2;
     std::map<TLMBusConnector*, int> busLogIds;
     std::map<Connector*, int> connectorLogIds;
+    std::string singleModel;
```
lochel: What is `singleModel` used for?

src/OMSimulatorLib/TLM/SystemTLM.cpp
```diff
@@ -419,6 +426,46 @@ oms_status_enu_t oms::SystemTLM::setPositionAndOrientation(const oms::ComRef &cr
...
+                                                 std::vector<std::string> &names,
+                                                 std::vector<int> &dimensions,
+                                                 std::vector<std::string> &domains)
+{
+    if(getComponents().find(cref) == getComponents().end()) {
+      return logError("Sub-model \""+std::string(cref)+"\" not found.");
+    }
+    singleModel = cref.c_str();
+    initialize();
+    singleModel = "";
```
lochel: Is `singleModel` any useful outside of the initialize method?

src/OMSimulatorLib/TLM/SystemTLM.h
robbr48: Interfaces are fetched by simulating the TLM model a single step. Since some submodels take much longer time to initialize than others, we don't want to always include all the submodels. With the `singleModel` argument, we instruct OMTLMSimulator to only include the specified submodel and no connections.

src/OMSimulatorLib/TLM/SystemTLM.cpp
robbr48: No, but `initialize` is a virtual function, so I cannot add an argument to it without bloating all the other classes with the same argument.

Issue #715, opened 2020-01-16
Saving the ssp again causes a crash


adeas31:
### Description

Saving the ssp once works fine but saving it again results in a crash.

### Steps to reproduce the behavior

Run the following script code,

```Lua
oms_setCommandLineOption("--suppressPath=true")
oms_setTempDirectory("./SSP-lua/")

function printStatus(status, expected)
  cmp = ""
  if status == expected then
    cmp = "correct"
  else
    cmp = "wrong"
  end

  if 0 == status then
    status = "ok"
  elseif 1 == status then
    status = "warning"
  elseif 3 == status then
    status = "error"
  end
  print("status:  [" .. cmp .. "] " .. status)
end

status = oms_newModel("model")
status = oms_addSystem("model.sc", oms_system_sc)

src, status = oms_list("model")
printStatus(status, 0)
print(src)

status= oms_export("model", "model.ssp")
printStatus(status, 0)

status = oms_export("model", "model.ssp")
printStatus(status, 0)
```

### Expected behavior

We should be able to save as many times as we want.

### Version and OS

- Version: [`OMSimulator v2.1.0-dev-147-g36ec2c7-mingw`]
- OS: [Win10, 64bit]

lochel:
@AnHeuermann Thanks for your fix.

@adeas31 Can we close this issue?

adeas31:
I tested a bit and it is working as expected. Thanks @AnHeuermann

PR #716, opened 2020-01-16
Overwrite files when copying with MINGW

AnHeuermann:
### Related Issues

Issue https://github.com/OpenModelica/OMSimulator/issues/715

### Purpose

Make `oms_copy_file` function consistent on MINGW and non MINGW systems

### Approach

Overwrite exisitng files when using `oms_copy_file`, so no error will be thrown.

### DIFF
--- a/testsuite/api/test_omsExport.py
+++ b/testsuite/api/test_omsExport.py
@@ -1,0 +1,65 @@
+## status: correct
+## teardown_command: rm -rf test_omsExport-py/
+## linux: yes
+## linux32: yes
+## mingw: yes
+## win: yes
+## mac: no
+
+from OMSimulator import OMSimulator
+oms = OMSimulator()
+
+oms.setCommandLineOption("--suppressPath=true")
+oms.setTempDirectory("./test_omsExport-py/")
+
+def printStatus(status, expected):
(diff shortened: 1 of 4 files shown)


testsuite/api/test_omsExport.py
```diff
@@ -0,0 +1,65 @@
...
+## linux32: yes
+## mingw: yes
+## win: yes
+## mac: yes
+
+from OMSimulator import OMSimulator
+oms = OMSimulator()
+
+oms.setCommandLineOption("--suppressPath=true")
+oms.setTempDirectory("./test_omsExport-lua/")
```
lochel: This should probably be `oms.setTempDirectory("./test_omsExport-py/")` but does not really matter I guess.

testsuite/api/test_omsExport.py
```diff
@@ -0,0 +1,65 @@
+## status: correct
+## teardown_command: rm -rf test_omsExport-lua/
+## linux: yes
+## linux32: yes
+## mingw: yes
+## win: yes
+## mac: yes
```
lochel: OMSimulatorPython isn't available on Mac OS X

PR #723, opened 2020-03-21
add support for SSP version 1.0

arun3688:
### Purpose
The purpose of this PR is to add support for SSP version 1.0 according to the specification.

### Approach
The main approach is support SSP version "1.0" and also support older version "Draft20180219", so that users can still import older version and run in OMSimulator. The main idea is to import the older version and export the older version to version 1.0.

### changes
The changes are maily done in importSSD_simulationInformation, as the <ssd:simulationInformation> is moved to vendor specific annotation <ssd:ssdAnnotations>, and from now on we export SSP according to version 1.0


arun3688:
@lochel  Please have a look into this PR

src/OMSimulatorLib/Connector.cpp
```diff
@@ -155,20 +193,21 @@ oms_status_enu_t oms::Connector::exportToSSD(pugi::xml_node &root) const
...
     break;
   case oms_signal_type_real:
-    node.append_attribute("type") = "Real";
+    node.append_child(oms::ssp::Version1_0::ssc::real_type);
     break;
   case oms_signal_type_string:
-    node.append_attribute("type") = "String";
+    node.append_child(oms::ssp::Version1_0::ssc::string_type);
     break;
+  //TODO handle Binary Types
```
lochel, line 210: What binary type? Please catch all unhandled cases and print an error.

arun3688: @ added log Error to catch the Binary types exceptions in connectors

src/OMSimulatorLib/Connector.cpp
```diff
@@ -136,6 +140,40 @@ oms::Connector* oms::Connector::NewConnector(const pugi::xml_node& node)
...
+    typeString = node.attribute("type").as_string();
+    return typeString;
+  }
+  else if (sspVersion == "1.0")
+  {
+    for(pugi::xml_node_iterator it = node.begin(); it != node.end(); ++it)
+    {
+      std::string name = it->name();
+      if (name == oms::ssp::Version1_0::ssc::real_type)
+        return typeString = "Real";
```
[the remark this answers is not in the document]
arun3688: @adeas31 You are correct, it is a typo fixed it

src/OMSimulatorLib/Connector.cpp
```diff
@@ -112,7 +112,12 @@ oms::Connector* oms::Connector::NewConnector(const pugi::xml_node& node, const s
     type = oms_signal_type_boolean;
   else if (typeString == "Enumeration")
     type = oms_signal_type_enum;
-  //TODO handle Binary Types
+  //TODO handle Binary Types for FMI-2.1 see specification
+  else if (typeString == "Binary")
```
lochel, line 116: just use an else to catch all unhandled cases

PR #725, opened 2020-04-14
export Parameter bindings to SSP

arun3688:
### Purpose
This PR, aims to export Parameters bindings to SSP.  With this functionality OMSimulator will now be able to support Parameter setting through SSps.

### Approach

Parameter.cpp   - Separate class which will be used to export and import Parameters both at the component and System level.


testsuite/OMSimulator/import_export_Parameters.lua
```diff
@@ -0,0 +1,144 @@
...
+
+-- instantiate FMUs
+oms_addSubModel("PI_Controller.co_sim.addP", "../resources/Modelica.Blocks.Math.Add.fmu")
+oms_addSubModel("PI_Controller.co_sim.P", "../resources/Modelica.Blocks.Math.Gain.fmu")
+oms_addSubModel("PI_Controller.co_sim.addI", "../resources/Modelica.Blocks.Math.Add3.fmu")
+
+-- setParameters
+oms_setReal("PI_Controller.co_sim.addP.k1", 10.0)
+oms_setReal("PI_Controller.co_sim.addP.k2", -1.0)
+oms_setReal("PI_Controller.co_sim.addI.k2",  2.0)
```
lochel, line 22: Please test also a start value of a non-parameter variable.

robha67:
Are parameter values stored in an SSD file or have you looked into the SSV format also?

arun3688:
@robha67 at the moment it is stored inline with SSD file.

lochel:
@robha67 As @arun3688 pointed out, we put all the parameters in the ssd file right now. However, we plan to extend this by using the ssv and ssm files to provide the full functionality, including common parameters on system level that get automatically assigned to different components as specified in the mapping file.

PR #729, opened 2020-04-29
import_export system level parameters in ssp

arun3688:
### Purpose

This PR will now able to support system level Parameter Binding in SSP.

### changed Files

System.cpp  : Binded the Real , Integer and Boolean start values to Parameter class

### DIFF
--- a/src/OMSimulatorLib/DirectedGraph.cpp
+++ b/src/OMSimulatorLib/DirectedGraph.cpp
@@ -263,7 +263,7 @@ void oms::DirectedGraph::calculateSortedConnections()
       Connector conA = nodes[edges[components[i][j]].first];
       Connector conB = nodes[edges[components[i][j]].second];
 
-      if (isValidConnection(conA, conB))
+      if (isValidConnection(conA.getName(), conB.getName(), conA, conB))
         SCC.push_back(std::pair<int, int>(edges[components[i][j]]));
     }
 
@@ -280,11 +280,8 @@ void oms::DirectedGraph::calculateSortedConnections()
 /*
  * Function which implements the allowed connections, according to SSP-1.0 connection table
  */
-bool oms::DirectedGraph::isValidConnection(const Connector& conA, const Connector& conB) const
+bool oms::DirectedGraph::isValidConnection(const ComRef& crefA, const ComRef& crefB, const Connector& conA, const Connector& conB) const
 {
-  const ComRef& crefA = conA.getName();
-  const ComRef& crefB = conB.getName();
-
   bool connectorA, connectorB;
 
   // Check connector A
--- a/src/OMSimulatorLib/DirectedGraph.cpp
+++ b/src/OMSimulatorLib/DirectedGraph.cpp
@@ -261,8 +261,8 @@ void oms::DirectedGraph::calculateSortedConnections()
     for (int j = 0; j < components[i].size(); ++j)
     {
       // flip causality checks for connectors (top-level crefs)
-      bool outA = nodes[edges[components[i][j]].first].getName().isValidIdent() ? nodes[edges[components[i][j]].first].isInput() : nodes[edges[components[i][j]].first].isOutput();
-      bool inB = nodes[edges[components[i][j]].second].getName().isValidIdent() ? nodes[edges[components[i][j]].second].isOutput() : nodes[edges[components[i][j]].second].isInput();
+      bool outA = nodes[edges[components[i][j]].first].getName().isValidIdent() ? nodes[edges[components[i][j]].first].isInput() : nodes[edges[components[i][j]].first].isOutput() || nodes[edges[components[i][j]].first].isParameter();
+      bool inB = nodes[edges[components[i][j]].second].getName().isValidIdent() ? nodes[edges[components[i][j]].second].isOutput() : nodes[edges[components[i][j]].second].isInput() || nodes[edges[components[i][j]].second].isParameter();
 
       if (outA && inB)
         SCC.push_back(std::pair<int, int>(edges[components[i][j]]));
--- a/src/OMSimulatorLib/DirectedGraph.cpp
+++ b/src/OMSimulatorLib/DirectedGraph.cpp
@@ -260,12 +260,14 @@ void oms::DirectedGraph::calculateSortedConnections()
     SCC.clear();
     for (int j = 0; j < components[i].size(); ++j)
     {
-      // flip causality checks for connectors (top-level crefs)
-      bool outA = nodes[edges[components[i][j]].first].getName().isValidIdent() ? nodes[edges[components[i][j]].first].isInput() : nodes[edges[components[i][j]].first].isOutput() || nodes[edges[components[i][j]].first].isParameter();
-      bool inB = nodes[edges[components[i][j]].second].getName().isValidIdent() ? nodes[edges[components[i][j]].second].isOutput() : nodes[edges[components[i][j]].second].isInput() || nodes[edges[components[i][j]].second].isParameter();
+      // is this check needed as we check all connections at topLevel, flip causality checks for connectors (top-level crefs)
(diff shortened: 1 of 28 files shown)


src/OMSimulatorLib/DirectedGraph.cpp
```diff
@@ -261,8 +261,8 @@ void oms::DirectedGraph::calculateSortedConnections()
     for (int j = 0; j < components[i].size(); ++j)
     {
       // flip causality checks for connectors (top-level crefs)
-      bool outA = nodes[edges[components[i][j]].first].getName().isValidIdent() ? nodes[edges[components[i][j]].first].isInput() : nodes[edges[components[i][j]].first].isOutput();
-      bool inB = nodes[edges[components[i][j]].second].getName().isValidIdent() ? nodes[edges[components[i][j]].second].isOutput() : nodes[edges[components[i][j]].second].isInput();
+      bool outA = nodes[edges[components[i][j]].first].getName().isValidIdent() ? nodes[edges[components[i][j]].first].isInput() : nodes[edges[components[i][j]].first].isOutput() || nodes[edges[components[i][j]].first].isParameter();
+      bool inB = nodes[edges[components[i][j]].second].getName().isValidIdent() ? nodes[edges[components[i][j]].second].isOutput() : nodes[edges[components[i][j]].second].isInput() || nodes[edges[components[i][j]].second].isParameter();
```
lochel, line 265: I still cannot remember what this does...

src/OMSimulatorLib/System.cpp
```diff
@@ -389,6 +393,12 @@ oms_status_enu_t oms::System::exportToSSD(pugi::xml_node& node) const
   if (oms_status_ok != element.getGeometry()->exportToSSD(node))
     return logError("export of system ElementGeometry failed");
 
+  // export top level system connectors
+  pugi::xml_node connectors_node = node.append_child(oms::ssp::Draft20180219::ssd::connectors);
```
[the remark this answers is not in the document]
arun3688: This is because, A top level system can have connectors and parameter bindings, that is why it is moved there, otherwise only subsystem connectors are exported

src/OMSimulatorLib/System.cpp
```diff
@@ -977,12 +993,16 @@ oms_status_enu_t oms::System::addConnection(const oms::ComRef& crefA, const oms:
   }
   else
   {
+    // TODO implement causality table which allows different possible connection according to specification
     if (!((conA->getCausality() == oms_causality_output && conB->getCausality() == oms_causality_input) ||
-      (conB->getCausality() == oms_causality_output && conA->getCausality() == oms_causality_input)))
+      (conB->getCausality() == oms_causality_output && conA->getCausality() == oms_causality_input) ||
+      (conA->getCausality() == oms_causality_parameter && conB->getCausality() == oms_causality_parameter)))
```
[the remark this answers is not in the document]
arun3688: This should be done

src/OMSimulatorLib/DirectedGraph.cpp
arun3688: This two lines are used to check whether the connection is valid or not, and isValidIdent() basically checks for whether it belongs to top level System or Elements

src/OMSimulatorLib/System.cpp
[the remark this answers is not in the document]
arun3688: All the valid combinations are now implemented

src/OMSimulatorLib/DirectedGraph.cpp
```diff
@@ -277,3 +282,43 @@ void oms::DirectedGraph::calculateSortedConnections()
 
   sortedConnectionsAreValid = true;
 }
+
+/*
+ * Function which implements the allowed connections, according to SSP-1.0 connection table
+ */
+bool oms::DirectedGraph::isValidConnection(const ComRef& crefA, const ComRef& crefB, Connector& conA, Connector& conB)
```
lochel, line 289: Why did you define `crefA` and `crefB` as an extra argument? This information is already present in `conA` and `conB`.

src/OMSimulatorLib/DirectedGraph.cpp
```diff
@@ -277,3 +282,43 @@ void oms::DirectedGraph::calculateSortedConnections()
...
+{
+  bool connectionA, connectionB;
+
+  // Check connection-A is valid
+  if (crefA.isValidIdent()) // smart way of detecting, top Level System
+    if (conA.isInput() || conA.isParameter()) // Top System level var's at ConnectionA can be input or parameter
+      connectionA = true;
+    else
+      connectionA = false;
```
lochel, line 298: Might even be easier to just write 
```C
connectionA = conA.isInput() || conA.isParameter(); // top-level vars of system must be inputs or parameters
```

src/OMSimulatorLib/DirectedGraph.cpp
```diff
@@ -277,3 +282,43 @@ void oms::DirectedGraph::calculateSortedConnections()
...
+
+/*
+ * Function which implements the allowed connections, according to SSP-1.0 connection table
+ */
+bool oms::DirectedGraph::isValidConnection(const ComRef& crefA, const ComRef& crefB, Connector& conA, Connector& conB)
+{
+  bool connectionA, connectionB;
+
+  // Check connection-A is valid
+  if (crefA.isValidIdent()) // smart way of detecting, top Level System
```
lochel, line 294: I would prefer to use brackets for those nested if conditions.

src/OMSimulatorLib/DirectedGraph.cpp
```diff
@@ -277,3 +282,43 @@ void oms::DirectedGraph::calculateSortedConnections()
...
+    // TODO check for inout, either FMI-2.0 or FMI-1.0 does not have inout
+    if (conB.isParameter() || conB.isInput()) // Element level vars at ConnectionB can be parameter or input, and inout
+      connectionB = true;
+    else
+      connectionB = false;
+
+  if (connectionA && connectionB)
+    return true;
+  else
+    return false;
```
lochel, line 322:
```C
return connectionA && connectionB; // both connectors must be valid in order to make the connection valid
```

src/OMSimulatorLib/DirectedGraph.cpp
```diff
@@ -277,3 +282,43 @@ void oms::DirectedGraph::calculateSortedConnections()
 
   sortedConnectionsAreValid = true;
 }
+
+/*
+ * Function which implements the allowed connections, according to SSP-1.0 connection table
+ */
+bool oms::DirectedGraph::isValidConnection(const ComRef& crefA, const ComRef& crefB, Connector& conA, Connector& conB)
+{
+  bool connectionA, connectionB;
```
lochel, line 291: This should probably be renamed to `connectorA` and `connectorB`.

src/OMSimulatorLib/System.cpp
```diff
@@ -2,8 +2,8 @@
  * This file is part of OpenModelica.
  *
  * Copyright (c) 1998-CurrentYear, Open Source Modelica Consortium (OSMC),
- * c/o Linköpings universitet, Department of Computer and Information Science,
- * SE-58183 Linköping, Sweden.
+ * c/o Link�pings universitet, Department of Computer and Information Science,
+ * SE-58183 Link�ping, Sweden.
```
lochel, line 6: What did you change here?

src/OMSimulatorLib/System.cpp
```diff
@@ -830,6 +850,11 @@ oms::Connector* oms::System::getConnector(const oms::ComRef& cref)
   return NULL;
 }
 
+std::string oms::System::getConnectorOwner(const oms::ComRef& cref)
```
lochel, line 853: Should we move this to `oms::ComRef`?

src/OMSimulatorLib/System.cpp
lochel: Encoding...

src/OMSimulatorLib/DirectedGraph.cpp
arun3688: Yes both must be valid to make a connection

src/OMSimulatorLib/DirectedGraph.cpp
arun3688: This is needed, because the crefs can only detect  if it is top level System or Element, the connectors.getName(), does not give the fullcrefs, it just has only the connectors name

Issue #742, opened 2020-06-03
Crash if you use Linux FMUs on Windows


adeas31:
### Description

OMSimulator crashes if by mistake you use Linux FMUs on Windows. This is not so critical on command line but is very bad for OMEdit since it crashes OMEdit as well.

The log file says,

```
Wed Jun 03 13:35:52 2020 | error:   [fmiLogger] module FMILIB: Could not change to the DLL directory C:/Users\adeas31\AppData\Local\Temp\OpenModelica\OMEdit\Test-f71gvox4\temp\0001_Gain\binaries\win64\
Wed Jun 03 13:35:52 2020 | error:   [fmiLogger] module FMILIB: System error: No error
Wed Jun 03 13:35:52 2020 | error:   [instantiate] Could not create the DLL loading mechanism (C-API). Error: System error: No error
```

### Steps to reproduce the behavior

Add the Linux FMUs to the model and try to instantiate the model.

### Expected behavior

An error message should be raised instead of a crash.

### Version and OS

- Version: `OMSimulator --version` OMSimulator v2.1.0-dev-167-g62a5731-mingw
- OS: Win10, 64bit

PR #743, opened 2020-06-03
Fix segmentation fault if shared object couldn't be loaded

lochel:
### Related Issues

* #742 

### Purpose

OMSimulator tries to unload a shared object even if it couldn't be loaded in the first place. This causes segmentation fault and OMEdit to crash.

### Approach

Unload the shared object only if is has been loaded in the first place 😀

### DIFF
--- a/src/OMSimulatorLib/ComponentFMUCS.cpp
+++ b/src/OMSimulatorLib/ComponentFMUCS.cpp
@@ -429,7 +429,7 @@ oms_status_enu_t oms::ComponentFMUCS::instantiate()
   // load the FMU shared library
   jmstatus = fmi2_import_create_dllfmu(fmu, fmi2_fmu_kind_cs, &callbackFunctions);
   if (jm_status_error == jmstatus)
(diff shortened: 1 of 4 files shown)


lochel:
OMEdit doesn't crash anymore and shows instead the following messages:

```log
[1] 14:49:05 Scripting Notification
Set working directory to "/tmp/OpenModelica_lochel/OMEdit"

[2] 14:49:11 Scripting Notification
New model "test" with corresponding temp directory "/tmp/OpenModelica_lochel/OMEdit/test-sesycw94"

[3] 14:49:27 Scripting Error
[fmiLogger] module FMILIB: Could not change to the DLL directory /tmp/OpenModelica_lochel/OMEdit/test-sesycw94/temp/0001_test/binaries/linux64/

[4] 14:49:27 Scripting Error
[fmiLogger] module FMILIB: System error: Success

[5] 14:49:27 Scripting Error
[instantiate] Could not load the FMU; it may be corrupted or may not support your platform
```

lochel:
I guess the error message should also point out the FMU that couldn't be loaded...

Issue #751, opened 2020-06-17
Error running non-OM SSP from command line


markaren:
### Description

I'm trying to load a .ssp (version 1.0) configuration using the OMSimulator command line. I have some issues. Now SSP 1.0 support is hard to come by and I'm not sure if OMSimulator suppports it? However, this SPP is valid in the two frameworks I've tested ([libcosim](https://github.com/open-simulation-platform/libcosim) and an still internal NTNU framework).

### Steps to reproduce the behavior

Run the attached SPP using OMSimulator

```
./OMSimulator.exe path/to/donfig.ssp
```

### Expected behavior

When `ssd:SystemStructureDescription name` includes the character "-" it will fail

E.g.
```
error:   [NewModel] "quarter-truck" is not a valid ident
Segmentation fault
```

Without the "-" in the name (attached file) another error appears:

```
error:   [NewSystem] internal error
```

### Version and OS

- Version:  v2.1.0-dev-172-g8995b12-win
- OS:  Win10, 64bit

[quarter-truck.zip](https://github.com/OpenModelica/OMSimulator/files/4792164/quarter-truck.zip)

lochel:
Thanks for the feedback. Identifiers in OMSimulator are rather restricted for now and must not contain hyphens.

markaren:
Just to add some more insight into this issue. I believe left out some important information I tried to run this as administrator in powershell as opposed to a bash shell as before. However, I still got the same error:

```
info:    Set temp directory to    "C:/Workspace"
info:    Set working directory to "C:/Workspace"
info:    New model "QuarterTruck" with corresponding temp directory "C:/Workspace\QuarterTruck-k0utd1kq"
info:    Set working directory to "C:/Workspace"
info:    Set working directory to "C:/Workspace\QuarterTruck-k0utd1kq"
error:   [NewSystem] internal error
info:    Set working directory to "C:/Workspace"
``` 

So it tries to create a folder `C:/Workspace\QuarterTruck-k0utd1kq`, but this folder is never created. I guess this is where the error lies.

markaren:
Ok, so I tried running this using the MINGW64 version on Windows. Here it succeeds to create the `QuarterTruck-XXXX` folder, however it only contains empty dummy folders and fails with the same error as above. It then fails to delete the generated folder, but at least that tells me that it manages to create the folder and on the win64 build it probably creates it and then deletes it afterwards during cleanup.

arun3688:
[...]

PR #752, opened 2020-06-23
Allow hyphens in crefs

lochel:
### Related Issues

* #751

### Purpose

OMSimulator is much stricter on crefs than the SSP specification. This PR will at least allow hyphens within crefs.


lochel:
I'll postpone this for now.

Issue #734, opened 2020-05-13
Remove Parameter Dependencies during Simulation 


arun3688:
### Purpose
At the moment , when connecting Top System Level parameters to Element level Parameters, the dependencies are analysed during initialization and simulation stage, But the parameter dependencies are not needed in Simulation and this should be removed.

lochel:
This is fixed 🎉

PR #757, opened 2020-07-17
Remove Parameter Dependencies during Simulation

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/734

### Purpose

This PR removes the top System Level Parameter dependencies during simulation stage, as it is already processed during initialization.

### Changed Files

System.cpp   - added a condition check , to filter the top level system parameters not to be included in dependency graph


src/OMSimulatorLib/System.cpp
```diff
@@ -1604,7 +1604,15 @@ oms_status_enu_t oms::System::updateDependencyGraphs()
       if (validConnection)
       {
         initialUnknownsGraph.addEdge(Connector(varA->getCausality(), varA->getType(), connection->getSignalA()), Connector(varB->getCausality(), varB->getType(), connection->getSignalB()));
-        outputsGraph.addEdge(Connector(varA->getCausality(), varA->getType(), connection->getSignalA()), Connector(varB->getCausality(), varB->getType(), connection->getSignalB()));
+
+        /*
+         * Remove top System level Parameters Dependencies during Simulation, as it is processed during initialization
+         * check for (TopLevel System and not parameter) (or) allow all element level connections
+         */
+        if ((connection->getSignalA().isValidIdent() && !varA->isParameter()) || !connection->getSignalA().isValidIdent())
```
lochel, line 1612: @arun3688 Why do you check for `connection->getSignalA().isValidIdent()`? Shouldn't it be enough to check for `variability != parameter?`

arun3688: yes we can check for only for variability != parameter?, see https://github.com/OpenModelica/OMSimulator/pull/757/commits/03b1c1e72c3deb529eb27083f9439ab06f9fe157

Issue #748, opened 2020-06-09
Deleting a connector in a system doesn't delete its connection


adeas31:
### Description

Deleting an element deletes the connection associated to it but deleting a connector doesn't delete the connection associated to it.

### Steps to reproduce the behavior

I have attached two cases. One is where element and its connection is deleted and we get the expected behavior and the other is where connector is deleted but its connection is left behind and the ssd becomes invalid.

[TestCases.zip](https://github.com/OpenModelica/OMSimulator/files/4751708/TestCases.zip)

### Expected behavior

The deletion of connector should also delete the connection.

### Version and OS

- Version: [`OMSimulator v2.1.0-dev-172-g8995b12-mingw`]
- OS: [Win10, 64bit]

arun3688:
deleting the connector should also delete the startValues associated with the connector along with the connections

lochel:
It works now 🎉

PR #758, opened 2020-07-22
delete a connector and its connection

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/748

### Purpose

This PR fixes the problem involved when deleting a connector , should also deletes it associated connections.

### changed files

System.cpp -- added new function "deleteConnectionsToConnector"  which deltes the connection associtated with the connector.

### DIFF
--- a/src/OMSimulatorLib/Component.h
+++ b/src/OMSimulatorLib/Component.h
@@ -101,6 +101,8 @@ namespace oms
     virtual oms_status_enu_t setInteger(const ComRef& cref, int value) {return logError_NotImplemented;}
     virtual oms_status_enu_t setReal(const ComRef& cref, double value) {return logError_NotImplemented;}
 
+    virtual oms_status_enu_t deleteStartValues(const ComRef& cref) {return oms_status_ok;}
+
     virtual oms_status_enu_t getRealOutputDerivative(const ComRef& cref, SignalDerivative& der) {return logError_NotImplemented;}
     virtual oms_status_enu_t setRealInputDerivative(const ComRef& cref, const SignalDerivative& der) {return logError_NotImplemented;}
 
--- a/src/OMSimulatorLib/Component.h
+++ b/src/OMSimulatorLib/Component.h
@@ -101,7 +101,7 @@ namespace oms
     virtual oms_status_enu_t setInteger(const ComRef& cref, int value) {return logError_NotImplemented;}
     virtual oms_status_enu_t setReal(const ComRef& cref, double value) {return logError_NotImplemented;}
 
-    virtual oms_status_enu_t deleteStartValues(const ComRef& cref) {return oms_status_ok;}
+    virtual oms_status_enu_t deleteStartValue(const ComRef& cref) {return oms_status_ok;}
 
     virtual oms_status_enu_t getRealOutputDerivative(const ComRef& cref, SignalDerivative& der) {return logError_NotImplemented;}
     virtual oms_status_enu_t setRealInputDerivative(const ComRef& cref, const SignalDerivative& der) {return logError_NotImplemented;}
--- a/src/OMSimulatorLib/Parameters.cpp
+++ b/src/OMSimulatorLib/Parameters.cpp
@@ -94,7 +94,7 @@ oms_status_enu_t oms::Parameters::deleteStartValue(const ComRef& cref)
     return oms_status_ok;
   }
 
-  return logWarning("failed to delete start value for signal \"" + std::string(cref) + "\"");
+  return oms_status_ok;
 }
 
 oms_status_enu_t oms::Parameters::exportToSSD(pugi::xml_node& node) const
--- a/src/OMSimulatorLib/Parameters.cpp
+++ b/src/OMSimulatorLib/Parameters.cpp
@@ -70,7 +70,7 @@ oms_status_enu_t oms::Parameters::setBoolean(const ComRef& cref, bool value)
   return oms_status_ok;
 }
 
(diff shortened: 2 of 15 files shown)


src/OMSimulatorLib/Component.h
```diff
@@ -101,6 +101,8 @@ namespace oms
     virtual oms_status_enu_t setInteger(const ComRef& cref, int value) {return logError_NotImplemented;}
     virtual oms_status_enu_t setReal(const ComRef& cref, double value) {return logError_NotImplemented;}
 
+    virtual oms_status_enu_t deleteStartValues(const ComRef& cref) {return oms_status_ok;}
```
lochel, line 104: This is just deleting a single start value at a time, right? Hence I would rename it to `deleteStartValue`.

src/OMSimulatorLib/System.cpp
```diff
@@ -1505,7 +1506,8 @@ oms_status_enu_t oms::System::delete_(const oms::ComRef& cref)
     for (int i=0; i<busconnectors.size()-1; ++i)
       if (busconnectors[i]->getName() == front)
       {
-        deleteAllConectionsTo(front);
+        //deleteAllConectionsTo(front);
```
lochel, line 1509: Why did you remove this call?

arun3688: i see two problems, 

1. This call actuall did not work, it did not delete the connections associated with the connector, and that is why this issue was repored
2. This call will delete all the connections assosciated with the "System or component", but that should not be the case , (e.g) we have two connection 

```
System2.C3 -> System1.C1
System2.C4 -> System1.C2
```
And now we want to delete 

`oms_delete("System2.C3")
`
The above call "deleteAllConectionsTo(front);" will delete all the connections associated with "System2" but that should not be the case , only this connection should be deleted

`System2.C3 -> System1.C1
`

src/OMSimulatorLib/Component.h
arun3688: renamed the function name to deleteStartValue see https://github.com/OpenModelica/OMSimulator/pull/758/commits/ee8d2f44ed637c3fcc93e87b8f81dbffd999b5ff

testsuite/OMSimulator/deleteConnector.lua
```diff
@@ -0,0 +1,241 @@
...
+oms_addConnection("deleteConnector.Root.System2.C4", "deleteConnector.Root.System1.C2")
+
+src = oms_list("deleteConnector")
+print(src)
+
+-- delete connector "deleteConnector.Root.System2.C3" and associated connection
+oms_delete("deleteConnector.Root.System2.C3")
+
+-- delete components connector and its startValue
+oms_delete("deleteConnector.Root.System1.Gain.k")
```
adeas31, line 36: Are we actually allowed to delete this, since `k` is part of the FMU?

arun3688: @adeas31 , yes we are allowed to do this,  as different tools does not export all the variables related to FMUs, some tools export only inputs and outputs

src/OMSimulatorLib/Parameters.cpp
```diff
@@ -70,6 +70,33 @@ oms_status_enu_t oms::Parameters::setBoolean(const ComRef& cref, bool value)
...
+  }
+  // booleanstartValues
+  auto boolValue = booleanStartValues.find(cref);
+  if (boolValue != booleanStartValues.end())
+  {
+    booleanStartValues.erase(boolValue);
+    return oms_status_ok;
+  }
+
+  return logError("deleteStartValue failed for : " + std::string(cref));
```
lochel, line 97: This should rather be a warning, right?

arun3688: you are correct, it can be a warning

src/OMSimulatorLib/System.cpp
```diff
@@ -1628,6 +1628,16 @@ oms_status_enu_t oms::System::delete_(const oms::ComRef& cref)
...
+      if (Flags::ExportParametersInline())
+      {
+        // delete startValues associated with the Connector
+        component->second->deleteStartValue(tail);
+      }
+      else
+      {
+        // delete startValues associated with the Connector from ssv file
+        component->second->deleteStartValue(cref);
+      }
```
lochel, line 1640: Please move this to `oms::Component::deleteConnector`

arun3688: moved the code to `oms::Component::deleteConnector` see https://github.com/OpenModelica/OMSimulator/pull/758/commits/b669e30afeea6d53ecfb6f68d9e9e71474db8dfc

src/OMSimulatorLib/System.cpp
```diff
@@ -1590,13 +1620,19 @@ oms_status_enu_t oms::System::delete_(const oms::ComRef& cref)
...
 
     auto component = components.find(front);
     if (component != components.end())
+    {
+      deleteAllConectionsTo(cref);
       return component->second->deleteConnector(tail);
+    }
   }
-
+  logWarning("failed to delete object \"" + std::string(getFullCref()+front) + "\"" + " because the identifier couldn't be resolved to any connector, component, system, or model");
```
lochel, line 1635: I guess it should be `std::string(getFullCref() + cref)`. `tail` doesn't necessarily need to be empty.

arun3688: changed the warning message https://github.com/OpenModelica/OMSimulator/pull/758/commits/6bf8b45f0071257e82813e5d62aa24321c2d47e8

PR #769, opened 2020-08-03
fix <ssd:annotation> tag to <ssc:annotation>

arun3688:
### Purpose

This PR fixes the bug in <ssd:annotation> which should be of type <ssc:annotation> according to SSP Specification

### example
 ```
<ssd:Annotations>
    <ssc:Annotation type="org.openmodelica">
        <oms:SimulationInformation>
            <oms:VariableStepSolver description="cvode" absoluteTolerance="0.000100" relativeTolerance="0.000100" minimumStepSize="0.000100" maximumStepSize="0.100000" initialStepSize="0.000100" />
        </oms:SimulationInformation>
    </ssc:Annotation>
</ssd:Annotations>
```


arun3688:
that is true, but the earlier version is a bug, do you want to support it ?

Issue #772, opened 2020-08-05
Install Python module OMSimulator with pip


AnHeuermann:
For the OpenModelica Python interface it is possible to install it with
```bash
$ pip3 install OMPython
```
It would be great to have the same option for OMSimulator Python interface.

Currently I can only use OMSimulator from Python when I run `OMSimulatorPython`.
Is there currently a way to install the module?
My Linter is complaining all the time `Unable to import 'OMSimulator'pylint(import-error)`.

superlou:
So, this is probably not the right way to do it, but to avoid needing to run `OMSimulatorPython .`, I set up my project like this by copying `libOMSimulator.so` and `OMSimulator.py` from the build of OMSimulator:

```
|-- oms/
|   |-- libOMSimulator.so   # I'm on Linux, would use the DLL on Windows
|   `--  OMSimulator.py
|-- main.py
```

With Python 3, I could then import OMS in `main.py` by:

```
from oms.OMSimulator import OMSimulator
```

I had to modify the line that loads the library in `OMSimulator.py` to match the new relative path: `self.obj=cdll.LoadLibrary("./oms/libOMSimulator.so")`. It would be better to use the `__file__` magic strings so that it isn't dependent on where you run the Python script from.

AnHeuermann:
@superlou Thanks, I did a similar thing to your suggestion:
```python
import sys
sys.path.append("/opt/OMSimulator-dev-181/lib")
from OMSimulator import OMSimulator
```
That's enough for me, but not very user friendly.

I needed to change OMSimulator.py as well. Also it depends on the OMSimulator.so location. That changes depending how you build OMSimulator (standalone or with OpenModelica) or download the artifacts from Jenkins.

arun3688:
@ setup file is available, package has to be uploaded to windows

arun3688:
@ this is implemented now with https://github.com/OpenModelica/OMSimulator-pip/pull/7, now we can use` pip install OMSimulator`

Issue #777, opened 2020-08-10
Add __enter__ and __exit__ to Python class


AnHeuermann:
## Problem
When using python module OMSimulator it would be nice to be able to exit a OMSimulator session.

It would be nice to use Pythons `with` statement:
```python
with OMSimulator() as oms:
  oms.getVersion()
```

**Priority**: Low (nice to have but not a real issue)

## Possible solution
Adding `__enter__` and `__exit__` to [OMSimulator.py.in](https://github.com/OpenModelica/OMSimulator/blob/master/src/OMSimulatorPython/OMSimulator.py.in) should do the trick.

superlou:
I started looking at this, and it's not as straightforward as I thought. My assumption is that we want the `with` statement to clean-up the simulation when terminated. To me, that would be calling `terminate` and `delete`. However, since those two functions require a cref. In the documentation I've seen, the typical process is to have a single model named `model`, with various sub-models underneath that root. If that's the case, the following is sufficient:

```
  def __enter__(self):
    return self.newModel('model')

  def __exit__(self, type, value, traceback):
    self.terminate('model')
    self.delete('model')
```

Alternatively,  we could have `__enter__` do nothing, and keep track of successfully added models each time `newModel` is called. Then the `__exit__` method will clean up all models. Which of these approaches is preferred?

The other thing I noticed is that `setTempDirectory` needs to be called before adding models. Because of this, the most straightforward way is to add a parameter to the constructor:

```
class OMSimulator:
  def __init__(self, temp_directory=None):
    self.obj=cdll.LoadLibrary("@OMSIMULATORLIB_STRING@")
    ...
    if temp_directory is not None:
        self.setTempDirectory(temp_directory)
```

That should be backwards compatible with everyone's usage of `OMSimulator.py`, but it's technically an API change, so I wanted to mention it.

superlou:
The model-tracking approach looks like this:

```python
class OMSimulator:
  def __init__(self, temp_directory=None):
    self.obj=cdll.LoadLibrary("@OMSIMULATORLIB_STRING@")

    ## oms_status_enu_t
    self.status_ok = 0
    self.status_warning = 1
    self.status_discard = 2
    self.status_error = 3
    self.status_fatal = 4
    self.status_pending = 5
    ...
    self.models = []

    if temp_directory is not None:
        self.setTempDirectory(temp_directory)

  def __enter__(self):
    return self

  def __exit__(self, type, value, traceback):
    for model in self.models:
[...]

PR #780, opened 2020-08-12
Fix load path for OMSimulator for Python

AnHeuermann:
### Related Issues
Running 
```python
>>> from OMSimulator import OMSimulator
>>> oms = OMSimulator()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/aheuermann1/OMSimulator-dev-158/lib/OMSimulator.py", line 8, in __init__
    self.obj=cdll.LoadLibrary("libOMSimulator.so")
  File "/usr/lib/python3.6/ctypes/__init__.py", line 426, in LoadLibrary
    return self._dlltype(name)
  File "/usr/lib/python3.6/ctypes/__init__.py", line 348, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libOMSimulator.so: cannot open shared object file: No such file or directory
>>> from OMSimulator import OMSimulator
>>> oms = OMSimulator()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/aheuermann1/OMSimulator-dev-158/lib/OMSimulator.py", line 8, in __init__
    self.obj=cdll.LoadLibrary(dllName)
  File "/usr/lib/python3.6/ctypes/__init__.py", line 426, in LoadLibrary
    return self._dlltype(name)
  File "/usr/lib/python3.6/ctypes/__init__.py", line 348, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libOMSimulator.so: cannot open shared object file: No such file or directory
```
can fail if:
-  OMSimulator was build toghether with OpenModelica (different folder structure)

or

-  The current directory is not the location where libOMSimulator.so is stored.

### Purpose
Make OMSimulator usable from Python without hacking our files.

### Approach
  - Load path gets set by CMake during build.
    This should work when building OMSimulator with and without
    OpenModelica.
  - It's now possible to give a custom dynamic library for OMSimulator.
    ```python
    >>> from OMSimulator import OMSimulator
    >>> oms = OMSimulator()
    >>> oms_specific_version = OMSimulator("/path/to/libOMSimulator.so")
    ```

Thanks @superlou.
I extended the way you changed your files described in https://github.com/OpenModelica/OMSimulator/issues/772.


superlou:
Thinking about steps to make the Python bindings more Pythonic like in #777, I was wondering if we see any other build-specific configuration steps in the Python file? Could we simplify the Python-binding development process by simply having the class first look for matching libs in its own directory, then see if someone has specified a `lib` parameter? That would remove the need to have any compile-time processing of the Python bindings. Since OMSimulator instances are made so rarely, there's no significant performance gain had by trying to guess the lib at compile time.

AnHeuermann:
I'm not sure if I understand you correctly. But depending on the way OMSimulator is build the lib will lie right next to it (if you build OMSimulator as a independent project) or in a subdirectory (if you build it with the OpenModelica superproject). So we need to add this in some sort to the loading.

And `self.obj=cdll.LoadLibrary(os.path.join(dirname, "@OMSIMULATORLIB_DIR_STRING@", "@OMSIMULATORLIB_STRING@"))` should contain the correct location of the dynamic C library after build. So there is no guessing.

But you can change this in #777 if you like. For now I would like to merge this PR.

superlou:
I agree with merging the PR as is (though I haven't tested it myself).

I was just thinking that down the road it might be nice to not need a compile step at all for the python code, i.e. removing `@OMSIMULATORLIB_DIR_STRING@` and `@OMSIMULATORLIB_STRING@` and using a little bit of Python logic to check for a lib file in the file directory.

AnHeuermann:
Ah, now I get it. Yes that would be possible.

lochel:
Thanks for the discussion and contribution. Feel free to come up with another PR for further improvements.

PR #782, opened 2020-08-12
Added methods to use OMSimulator in "with" statement

superlou:
### Related Issues

This addresses #777.

### Purpose

Improve Python ergonomics for the `OMSimulator` class.

### Approach

This PR provides the magic `__enter__` and `__exit__` methods so that an `OMSimulator` instance can be used in a `with` statement as follows:

```python
from OMSimulator import OMSimulator

with OMSimulator('./temp') as oms:
    oms.newModel("model")
    oms.addSystem("model.root", oms.system_sc)

    # instantiate FMUs
    oms.addSubModel("model.root.system1", "FMUs/System1.fmu")
    oms.addSubModel("model.root.system2", "FMUs/System2.fmu")

    # add connections
    oms.addConnection("model.root.system1.y", "model.root.system2.u")
    oms.addConnection("model.root.system2.y", "model.root.system1.u")

    # simulation settings
    oms.setResultFile("model", "results.mat")
    oms.setStopTime("model", 0.1)
    oms.setFixedStepSize("model.root", 1e-4)

    oms.instantiate("model")
    oms.setReal("model.root.system1.x_start", 2.5)

    oms.initialize("model")
    oms.simulate("model")
```

When the `with` statement completes, or is otherwise terminated, the `terminate` and `delete` methods are called for all models that have been created. Currently, the models are tracked in the Python side, but if there is a way to get them from the API, that might be a more robust way to determine all models that need cleanup.

Since the temp directory is needed immediately, it can be passed as an argument to the constructor.


lochel:
I am a bit reluctant because it would still be possible to change the global scope outside the with scope. If possible, it would be better to load and unload the OMSimulator completely.

superlou:
I tentatively agree with @lochel on preferring to release the library resources. However, I haven't found a method that isn't platform specific or hacky. Even if we unload during `__exit__`, couldn't someone change the global scope in another process? Can you provide an example of how someone might accidentally misuse the current implementation? Maybe there's a Python-side way to prevent misuse.

lochel:
I cannot come up with an example using the available set of python bindings. However, the OMSimulator API provides functions to set global settings. Those might be problematic.
Anyways, I am fine with merging this PR for now, since it actually works rather fine.

PR #784, opened 2020-08-13
Removing OMSysIdent

lochel:
We have to drop OMSysIdent for now, because it slows down the development and isn't actually maintained. However, we can bring it back if needed.

I will keep this open until the next dev meeting. Until then we can start a discussion here.


bernhard-thiele:
Well, I find it a bit sad to remove it, though I understand your maintenance argument. Unfortunately, I didn't have the time to follow the OMSimulator development as closely as it would be needed for bringing in enhancements to this module (though I'm also not aware of issues which needed urgent fixing).

I think that it would be possible to separate the OMSysIdent code and its dependencies from the rest of OMSimulator in its own project. Possibly one would need to extend the public OMSimulator API a bit, but after that one could use OMSimulator as a dependency to the new OMSysIdent project. Hence, OMSysIdent would only do calls to the public OMSimulator DLL API. This would also ease the maintenance of OMSysIdent, since one doesn't need to follow and understand internal OMSimulator changes. However, it would be quite some initial work to set it up. Particularly, if one has the ambition of getting CI and tests for several platforms as well as an end user friendly integration with supported scripting languages (though one could cut some support for easing the maintenance burden and the project set-up). However, last but not least the OMSysIdent project would still need a maintainer. This still might be the most promising approach if one wants to preserve the functionality...

lochel:
@bernhard-thiele I agree with you. It would be a shame to lose it, but at the same time it shouldn't interfere with the maintained part of OMSimulator. Therefore, we agreed in today’s meeting on making this a separate project that uses only the public API of OMSimulator. We will probably drop the Lua interface and only continue with the Python interface.
 
@sjoelund volunteered to create a new repository preserving the history of all the commits. The main sources are located in src/OMSysIdentLib and the python bindings are located in src/OMSimulatorPython.

sjoelund:
https://github.com/OpenModelica/OMSysIdent was created using `git filter-repo --path src/OMSimulatorPython/ --path src/OMSysIdentLib/`

lochel:
The next steps would be to move ceres and the other dependencies to the new repository and to work on the build system. To make things a bit easier we should probably just focus on python and linux for now. @bernhard-thiele What do you think about the plans?

lochel:
This was not supposed to be merged yet, but I think it would have anyways been merged eventually. @bernhard-thiele I hope you like the approach that we chose.

bernhard-thiele:
@lochel Fine for me. I'll experiment with the project the next days and will try to get it running. Focusing on Linux and Python first is a good approach. This is also an opportunity to update the dependencies to the latest Ceres Solver.

Issue #776, opened 2020-08-10
Support non-real connection between CSV input and FMU


AnHeuermann:
## Description
When connecting an non-real input (boolean, integer) from a CSV to set as input to an FMU OMSimulator will issue errors and not connect the variables.

### Example Feedthrough.fmu
See e.g. the example FMU `Feedthrough` from https://github.com/modelica/fmi-cross-check/tree/master/fmus/2.0/cs/linux64/Test-FMUs/0.0.2/Feedthrough

```bash
$ OMSimulator \
    --deleteTempFiles=true \
    --stripRoot=true \
    --intervals=1500 \
    --suppressPath=true \
    --timeout=60 \
    Feedthrough.lua
```

```lua
-- Lua file for Feedthrough.fmu
oms_setTempDirectory("temp")
oms_newModel("model")
oms_addSystem("model.root", oms_system_wc)

-- instantiate FMU
oms_addSubModel("model.root.input", "Feedthrough_in.csv")
oms_addSubModel("model.root.fmu", "/Feedthrough.fmu")

-- Connect inputs to FMU
oms_addConnection("model.root.input.real_tunable_param", "model.root.fmu.real_tunable_param")
oms_addConnection("model.root.input.real_continuous_in", "model.root.fmu.real_continuous_in")
oms_addConnection("model.root.input.real_discrete_in", "model.root.fmu.real_discrete_in")
oms_addConnection("model.root.input.int_in", "model.root.fmu.int_in")
oms_addConnection("model.root.input.bool_in", "model.root.fmu.bool_in")

-- Simulation settings
oms_setResultFile("model", "Feedthrough_out.csv")
oms_setStartTime("model", 0.0)
oms_setStopTime("model", 2.0)
oms_setTolerance("model", 1e-05)

-- Instantiate, initialize and simulate
oms_instantiate("model")
oms_initialize("model")
oms_simulate("model")
oms_terminate("model")
oms_delete("model")
```

```bash
error:   [addConnection] Type mismatch in connection: input.int_in -> fmu.int_in
error:   [addConnection] Type mismatch in connection: input.bool_in -> fmu.bool_in
```

## Possible solution
If I understand it correctly all FMU inputs are always real.
So a typecast to real would be perfectly fine, as long booleans are still treated as booleans.
Using `setInteger` or `setBoolean` depending on variable type could be a solution as well.

## Alternative solutions
Inputs for FMUs can be non-real as well. Then everything should work as expected.

[...]

PR #785, opened 2020-08-13
allow non real connections from csv input

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/776

### Purpose
This PR allows non Real connection from csv inputs to fmus, 

### Approach

System.cpp - Allow connection from csv inputs which are of non real inputs
ComponentTable.cpp - add `getInteger() `and `getBoolean()` methods

### DIFF
--- a/src/OMSimulatorLib/ComponentTable.cpp
+++ b/src/OMSimulatorLib/ComponentTable.cpp
@@ -226,18 +226,24 @@ oms_status_enu_t oms::ComponentTable::getInteger(const oms::ComRef& cref, int& v
 
   for (int i=0; i<series[cref]->length; ++i)
   {
-    if (time == series[cref]->time[i])
+    if(i == 0 && time < series[cref]->time[i])
+    {
+      return logError("invalid csv input series (cref=" + std::string(cref) + ", time=" + std::to_string(time) + " cannot be less than first time point in csv " + std::to_string(series[cref]->time[i]) + ")");
+    }
+    else if(time == series[cref]->time[i])
     {
       value = series[cref]->value[i];
       return oms_status_ok;
     }
-    else if(time < series[cref]->time[i])
+    else if (time < series[cref]->time[i])
     {
       value = series[cref]->value[i-1];
       return oms_status_ok;
     }
   }
+
   value = 0;
+  return oms_status_ok;
   return logError("out of range (cref=" + std::string(cref) + ", time=" + std::to_string(time) + ")");
 }
 
@@ -251,7 +257,11 @@ oms_status_enu_t oms::ComponentTable::getBoolean(const oms::ComRef& cref, bool& 
 
   for (int i=0; i<series[cref]->length; ++i)
   {
-    if (time == series[cref]->time[i])
+    if(i == 0 && time < series[cref]->time[i])
+    {
+      return logError("invalid csv input series (cref=" + std::string(cref) + ", time=" + std::to_string(time) + " cannot be less than first time point in csv " + std::to_string(series[cref]->time[i]) + ")");
+    }
+    else if (time == series[cref]->time[i] )
     {
       value = series[cref]->value[i];
       return oms_status_ok;
--- a/src/OMSimulatorLib/ComponentTable.cpp
+++ b/src/OMSimulatorLib/ComponentTable.cpp
@@ -216,6 +216,56 @@ oms_status_enu_t oms::ComponentTable::getReal(const oms::ComRef& cref, double& v
   return logError("out of range (cref=" + std::string(cref) + ", time=" + std::to_string(time) + ")");
 }
 
+oms_status_enu_t oms::ComponentTable::getInteger(const oms::ComRef& cref, int& value)
+{
+  if (!resultReader)
+    logError("the table isn't initialized properly");
+
+  if (series.find(cref) == series.end())
+    series[cref] = resultReader->getSeries(cref.c_str());
+
+  for (int i=1; i<series[cref]->length; ++i)
+  {
+    if (series[cref]->time[i-1] == time)
+    {
+      value = series[cref]->value[i-1];
+      return oms_status_ok;
+    }
+    else if (series[cref]->time[i-1] <= time && series[cref]->time[i] >= time)
+    {
+      value = series[cref]->value[i-1];
+      return oms_status_ok;
+    }
+  }
+  value = 0;
+  return logError("out of range (cref=" + std::string(cref) + ", time=" + std::to_string(time) + ")");
+}
+
+oms_status_enu_t oms::ComponentTable::getBoolean(const oms::ComRef& cref, bool& value)
+{
+  if (!resultReader)
+    logError("the table isn't initialized properly");
+
+  if (series.find(cref) == series.end())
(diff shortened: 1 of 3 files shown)


src/OMSimulatorLib/ComponentTable.cpp
```diff
@@ -216,6 +216,56 @@ oms_status_enu_t oms::ComponentTable::getReal(const oms::ComRef& cref, double& v
...
+  for (int i=1; i<series[cref]->length; ++i)
+  {
+    if (series[cref]->time[i-1] == time)
+    {
+      value = series[cref]->value[i-1];
+      return oms_status_ok;
+    }
+    else if (series[cref]->time[i-1] <= time && series[cref]->time[i] >= time)
+    {
+      value = series[cref]->value[i-1];
```
AnHeuermann, line 236: Interpolating an integer is not a good idea. They are only allowed to change at intervals, so returning the last computed one is always correct.

Is my understanding correct?

arun3688: yes that is correct, and this what i discussed with Lennart

lochel:
> The outputs on the other side are fishy, but that is a different story and not related to this PR.

@AnHeuermann What do you mean?

src/OMSimulatorLib/ComponentTable.cpp
```diff
@@ -216,6 +216,56 @@ oms_status_enu_t oms::ComponentTable::getReal(const oms::ComRef& cref, double& v
...
+      return oms_status_ok;
+    }
+  }
```
lochel, line 264: I think this logic is wrong for the last point of a table. It only works if using interpolation as we do for the implementation of continuous signals.

lochel:
We also must describe this behavior in the user's guide: Interpolation for continuous signals and "zero-order hold" for discrete signals.

src/OMSimulatorLib/ComponentTable.cpp
arun3688: @lochel Yes you are correct , the last point of a table will not work

AnHeuermann: Also the return value is not different for the two if cases. So one could replace it by one if-statement.
```c
for (int i=1; i<series[cref]->length; ++i)
{
    if (series[cref]->time[i-1] <= time && series[cref]->time[i] >= time)
    {
        value = series[cref]->value[i-1];
        return oms_status_ok;
    }
}
```

AnHeuermann:
## I had a closer look at [Feedthrough.fmu](https://github.com/modelica/fmi-cross-check/tree/master/fmus/2.0/cs/linux64/Test-FMUs/0.0.2/Feedthrough) from [FMI Cross-Check](https://github.com/modelica/fmi-cross-check)  that has a number of inputs and will output them 1 to 1.

[Feedthrough_in.csv](https://github.com/modelica/fmi-cross-check/blob/master/fmus/2.0/cs/linux64/Test-FMUs/0.0.2/Feedthrough/Feedthrough_in.csv):
```CSV
time,real_tunable_param,real_continuous_in,real_discrete_in,int_in,bool_in
0.0,0,0.0,0,0,0
0.5,0,0.0,0,0,0
0.5,0,2.0,0,0,0
1.0,0,1.0,1,1,1
1.5,-1,1.0,1,1,1
2.0,-1,1.0,1,1,1
```

I have connected the CSV input with the FMU:
Lua file:
```lua
-- Lua file for Feedthrough.fmu
oms_setTempDirectory("/temp")
oms_newModel("model")
oms_addSystem("model.root", oms_system_wc)

-- instantiate FMU
oms_addSubModel("model.root.input", "Feedthrough_in.csv")
oms_addSubModel("model.root.fmu", "Feedthrough.fmu")

-- Connect inputs to FMU
oms_addConnection("model.root.input.real_tunable_param", "model.root.fmu.real_tunable_param")
oms_addConnection("model.root.input.real_continuous_in", "model.root.fmu.real_continuous_in")
oms_addConnection("model.root.input.real_discrete_in", "model.root.fmu.real_discrete_in")
oms_addConnection("model.root.input.int_in", "model.root.fmu.int_in")
oms_addConnection("model.root.input.bool_in", "model.root.fmu.bool_in")

-- Simulation settings
oms_setResultFile("model", "Feedthrough_out.csv")
oms_setStartTime("model", 0.0)
oms_setStopTime("model", 2.0)
oms_setTolerance("model", 1e-05)
initialStepSize, minimumStepSize, maximumStepSize, status = oms_getVariableStepSize("model")
oms_setVariableStepSize("model", 0.004, minimumStepSize, 0.004)
oms_setFixedStepSize("model", 0.004)

-- Instantiate, initialize and simulate
oms_instantiate("model")
oms_initialize("model")
oms_simulate("model")
oms_terminate("model")
oms_delete("model")
```

## Boolean input:
We are missing the change at `time=1.0`:
image
In blue: `input.bool_in`
In red: `fmu.bool_in`

Then of course the output is wrong as well. Compare with the [reference result](https://github.com/modelica/fmi-cross-check/blob/master/fmus/2.0/cs/linux64/Test-FMUs/0.0.2/Feedthrough/Feedthrough_ref.csv).

**But can this even work?**
Looking at the source code of the FMU reveals that there is some string magic going on:
```c
Status getBoolean(ModelInstance* comp, ValueReference vr, bool *value, size_t *index) {
    calculateValues(comp);
    switch (vr) {
        case vr_bool_in:
            value[(*index)++] = M(boolean);
            return OK;
        case vr_bool_out:
            value[(*index)++] = M(boolean) && (strcmp(M(string), "FMI is awesome!") == 0);
            return OK;
        default: return Error;
    }
}
```
But the internal variable `M(string)` a.k.a. `string_param` with start value `"Set me!"` will not be changed by any tool unless done manually.

## Integer input:
The CSV file is changing the value at `time=1.0` and for `input.int_in` we get this but don't get it to change the output `fmu.int_in`
image

## Download:
Result from OMSimulator: [Feedthrough_out.zip](https://github.com/OpenModelica/OMSimulator/files/5073805/Feedthrough_out.zip)

arun3688:
@AnHeuermann , I plotted the mat file in OMEdit and i get correct results, I think OMEdit has some problems i guess, when i first plotted i also got the same results like you and then i plotted in a new window and i got this result, can you try it again
boolout

AnHeuermann:
You are right. Now it is working for `bool_in` and `int_in`. I don't get why it didn't show the correct result on the first try.
But for `real_tunable_param` I still get a wrong result.
image

But we can fix that problem with the real tunable parameter in a different PR.

Also I discussed with @lochel that we can create a OpenModelica FMU that does this input=output with all sorts of inputs.
Maybe adding that to OMSimulator/testsuite or OPenMOdelica/testsuite is a good idea.

src/OMSimulatorLib/ComponentTable.cpp
arun3688: @lochel  this problem should be fixed by this https://github.com/OpenModelica/OMSimulator/pull/785/commits/07ca3bfb305a88ed9ee8878525df6c0d5989da8f

Issue #749, opened 2020-06-12
Instantiation fails after a call to oms_loadSnapshot


adeas31:
### Description

Once a call to `oms_loadSnapshot` is made then the `oms_instantiate` produces weird errors and fails.

### Steps to reproduce the behavior

Use the attached ssp and the lua script to reproduce the error. If you comment the lines 31 and 32 in the script then everything works fine.

[Test.zip](https://github.com/OpenModelica/OMSimulator/files/4770108/Test.zip)

### Expected behavior

Instantiation should not fail due to `oms_loadSnapshot` calls.

### Version and OS

- Version: [`OMSimulator v2.1.0-dev-172-g8995b12-mingw`]
- OS: [Win10, 64bit]

lochel:
This is a slightly simpler test case, which is nicer to debug because it doesn't contain an ssp or fmu file:

```lua
oms_newModel("test")
oms_addSystem("test.root", oms_system_wc)
oms_addSubModel("test.root.A", "test.csv")

src1 = oms_list("test")
oms_setStopTime("test", 10)

oms_loadSnapshot("test", src1)
oms_instantiate("test")
oms_initialize("test")
oms_terminate("test")
oms_delete("test")
```

test.csv can be whatever, e.g.
```csv
time,signal
0.0,0
10.0,10
```

arun3688:
The above lua script should work with this https://github.com/OpenModelica/OMSimulator/pull/799, But still other problem exist when calling oms_loadsnapshot if fmus are involved, see below lua script 

```
oms_newModel("test")
oms_addSystem("test.root", oms_system_wc)
oms_addSubModel("test.root.add", "../resources/Modelica.Blocks.Math.Add.fmu")
oms_addSubModel("test.root.gain", "../resources/Modelica.Blocks.Math.Gain.fmu")
oms_addConnection("test.root.add.y", "test.root.gain.u")
src1 = oms_list("test")
oms_setStopTime("test", 10)
oms_loadSnapshot("test", src1)
oms_instantiate("test")
oms_initialize("test")
oms_terminate("test")
oms_delete("test")
```
The following errors are reported 

```
error:   [fmiLogger] module FMILIB: Could not change to the DLL directory C:/OPENMODELICAGIT\OpenModelica\OMSimulator\testsuite\Test\loadSnapchat\test-8adxfybg\temp\0001_add\binaries\win64\
error:   [fmiLogger] module FMILIB: The FMU contains no binary for this platform.

```

[...]

PR #799, opened 2020-09-01
use absolute path to CSV Reader

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/749

### Purpose

This PR should fix  oms_loadSnapshot called before instantiation.


lochel (approved):
Nice fix!

Issue #771, opened 2020-08-05
API to delete startValues


arun3688:
**Is your feature request related to a problem? Please describe.**
https://github.com/OpenModelica/OMSimulator/issues/748

**Describe the solution you'd like**
The current OMSimulator API `oms_delete()`  will delete models, connectors and its connections and its startValues, But we want to delete StartValues alone , Consider the below example

```
<ssd:Connectors>
    <ssd:Connector name="C3" kind="output">
        <ssc:Real />
    </ssd:Connector>
</ssd:Connectors>
<ssd:ParameterBindings>
    <ssd:ParameterBinding>
        <ssd:ParameterValues>
            <ssv:ParameterSet version="1.0" name="parameters">
                    <ssv:Parameter name="C3">
                        <ssv:Real value="20" />
                    </ssv:Parameter>
                </ssv:Parameters>
            </ssv:ParameterSet>
        </ssd:ParameterValues>
    </ssd:ParameterBinding>
</ssd:ParameterBindings>
```
The current API oms_delete("C3"), will delete Connector and its startValue,  But we want to delete only start value

**Describe alternatives you've considered**
We need to have to an separate API ` oms_deleteStartValues("C3")` or extend the current API `oms_delete` with suffix `start` added to the end some thing like below

`oms_delete('"C3:start")`

What should be a good option ?

adeas31:
Can't we use `oms_setReal`?

arun3688:
`oms_setReal()` sets the value,  but we want to delete it, Do you mean to set the value to `0 `using `oms_setReal("C3", "0") ` ?

lochel:
I am in favor of `oms_delete("C3:start")` because it doesn't require a new API call. I think we have already too many.

adeas31:
> `oms_setReal()` sets the value, but we want to delete it, Do you mean to set the value to `0 `using `oms_setReal("C3", "0") ` ?

Yes. `0` would be anyway assumed default value, rigth?

lochel:
No, it depends on the actual start values in the FMUs.

adeas31:
I mean if there is not actual start value in the FMU.

lochel:
This is implemented in #804

PR #804, opened 2020-09-03
delete startvalues with suffix :start

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/771
### Purpose

This PR deletes the startvalue alone when using `oms_delete` with suffix `:start`

### Example
`oms_delete("deleteStartValues.Root.C3:start")`

### DIFF
--- a/src/OMSimulatorLib/System.cpp
+++ b/src/OMSimulatorLib/System.cpp
@@ -1566,13 +1566,8 @@ oms_status_enu_t oms::System::delete_(const oms::ComRef& cref)
     // check cref has ":start" suffix at the end to delete only start values
     if (front.hasSuffixStart())
     {
-      // add prefix to cref start values if ssv file is used (e.g) f:start -> foo.f:start
-      if(!Flags::ExportParametersInline())
-      {
-        front = getCref()+front;
-      }
-      if (oms_status_ok != startValues.deleteStartValue(front))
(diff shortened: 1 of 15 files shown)


src/OMSimulatorLib/System.cpp
```diff
@@ -1669,6 +1677,40 @@ oms_status_enu_t oms::System::delete_(const oms::ComRef& cref)
...
+oms_status_enu_t oms::System::deleteStartValue(const ComRef& cref)
+{
+  oms::ComRef tail(cref);
+  oms::ComRef front = tail.pop_front();
+
+  if (tail.isEmpty())
+  {
+    if (oms_status_ok != startValues.deleteStartValue(front))
+    {
+      logWarning("failed to delete start value \"" + std::string(getFullCref()+cref)+":start" + "\"" + " because the identifier couldn't be resolved to any System");
```
lochel, line 1689: typo

lochel: and should we really return `oms_status_ok`?

arun3688: Yes we need to return oms_status_ok, because, otherwise the logWarning at the end of function 
 ` logWarning("failed to delete start value \"" + std::string(getFullCref()+cref)+":start" + "\"" + " because the identifier couldn't be resolved to any connector, component, system, or model");
`
will be printed for all the values, and this warning should be printed only for `unknown ` system or component which you can see in testcase

lochel (approved):

Issue #768, opened 2020-08-02
Documentation request on modifying the simulation after initialization


superlou:
**Is your feature request related to a problem? Please describe.**

I am trying to make a real-time hardware-in-the-loop simulation using `stepUntil` to move the simulation forward periodically while taking external values into the simulation after each 100 step. My proof-of-concept attempt uses Python to try to change the voltage on a power supply to a lower value after 1.0 second. However, I don't see the expected reduced current 10 seconds later.

```
#!/usr/bin/python3
from oms.OMSimulator import OMSimulator

def main():
    oms = OMSimulator()
    oms.setTempDirectory('./temp/')
    oms.newModel('model')
    oms.addSystem('model.root', oms.system_sc)

    oms.addSubModel('model.root.system1', './segfault_debug/SimpleElectrical.fmu')
    oms.setResultFile('model', '')
    oms.instantiate('model')
    oms.initialize('model')

    oms.stepUntil('model', 1.0)
    print(oms.getReal('model.root.system1.constantVoltage.i'))

    oms.setReal('model.root.system1.constantVoltage.V', 10.0)
    oms.stepUntil('model', 10.0)
    print(oms.getReal('model.root.system1.constantVoltage.i'))

    oms.terminate('model')
    oms.delete('model')


if __name__ == '__main__':
    main()
```

![image](https://user-images.githubusercontent.com/709695/89131412-8d7d9c80-d4da-11ea-9d1a-aa1ed9552027.png)

**Describe the solution you'd like**

Per the `setReal` [documentation](https://openmodelica.org/doc/OMSimulator/master/html/OMSimulatorPython.html#setreal), it looks like there are restrictions on updating properties after initialization. There is guidance that after initialization `setReal` can be used to force external inputs. Is it possible to add a description of how you would go about modifying signals after initialization? I might also be misunderstanding what is an "external input." Are there other limitations?

If someone can help me understand the right way to approach this, I can take a shot at writing up documentation for it.

**Describe alternatives you've considered**
I've tried modifying different properties and signals without any consistent reaction during the simulation.

superlou:
I think I might have just been really dense. Does "external inputs" refer to interfaces? I'm getting pretty reasonable inputs (for step changes) if my FMU has interfaces:

![image](https://user-images.githubusercontent.com/709695/89230599-9a20f400-d5b1-11ea-93e5-526c8b19bd95.png)

However, `setBoolean` always results in the following error:

```
[...]

PR #807, opened 2020-09-09
set and get values and startValues

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/768 and also for OMEdit 

### Purpose

This PR allows to set and get values and start values.

### Approach
use  suffix `:start ` to set start Values before instantiatiaon and  without suffix to set values after initialization
 
Example
```
oms_setReal("setAndGetStartValues.Root.C1:start", 20.0) //  with suffix :start , allowed before instantiation 
oms_instantiate("setAndGetStartValues")
oms_initialize("setAndGetStartValues")
oms_setReal("setAndGetStartValues.Root.C1", 50.0) // without suffix :start allowed only after initialization

```

### DIFF
--- a/src/OMSimulatorLib/ComponentFMUCS.cpp
+++ b/src/OMSimulatorLib/ComponentFMUCS.cpp
@@ -444,22 +444,22 @@ oms_status_enu_t oms::ComponentFMUCS::instantiate()
     return logError_FMUCall("fmi2_import_instantiate", this);
 
   // set start values
-  for (const auto& v : startValues.booleanStartValues)
+  for (const auto& v : startValues.booleanValues)
   {
     oms::ComRef cref = getValidCref(v.first);
-    if (oms_status_ok != setBoolean(cref, v.second))
+    if (oms_status_ok != setBoolean(cref, std::get<1>(v.second).getValue()))
       return logError("Failed to set start value for " + std::string(v.first));
   }
-  for (const auto& v : startValues.integerStartValues)
+  for (const auto& v : startValues.integerValues)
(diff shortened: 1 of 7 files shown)


superlou:
Is this purely about the OMEdit GUI, or is it possible this helps with #768?

arun3688:
@superlou hi this is for both scripting and OMEdit and thanks for pointing out the issue you have mentioned https://github.com/OpenModelica/OMSimulator/issues/768,  it should fix that as well

superlou:
@arun3688, thanks for the clarification. So I should be able to arbitrarily set values after initializing and between `oms_stepUntil(...)` calls? I'm going to try it out on my system, but I'm unfortunately away from my home computer until at least next week.

When setting boolean values after initialization previously, I would get `fmi2SetBoolean: Illegal call sequence. fmi2SetBoolean is not allowed in Continuous-Time Mode state.` Should that also be resolved, or is that fundamentally a different issue dealing with how the simulation is run?

adeas31:
@arun3688 do you want me to test this with OMEdit or will you do this yourself?

lochel:
@adeas31 No, you shouldn't test it yet. This is a fairly early version of the PR and far from working.

arun3688:
@adeas31 you can use the current OMSimulator master to set the start values in OMEdit gui, You can use the normal `getReal() ` API to get the start values and set it see https://github.com/OpenModelica/OMSimulator/pull/810,  You don't need to use any kind of `:start` suffix as mentioned  in this PR

Issue #812, opened 2020-09-18
Better non-linear solver for algebraic loops between components.


AnHeuermann:
**Is your feature request related to a problem? Please describe.**
Related to #808.

**Describe the solution you'd like**
Adding kinsol as alternative non-linear solver should be appropriate. It's already inside 3rdParty.

lochel:
@AnHeuermann What is the status?

AnHeuermann:
Kinsol was added with c8ec782983acc160271c0368afeb4b70ba861463. We need to verify and test some more, but this issue can be closed.

PR #816, opened 2020-09-24
Update 3rdParty with newer KINSOL/CVODE

AnHeuermann:
### Related Issues
Preparation for https://github.com/OpenModelica/OMSimulator/issues/812

### Purpose

When adding KINSOL we should use a newer version than 2.7.0

### Approach

Updated 3rdParty


AnHeuermann:
Mixing KINSOL and CVODE with different versions is not working, so I updated CVODE as well.

lochel (approved):

PR #826, opened 2020-10-05
Make error in broken dependencies from modelDescription.xml fatal

AnHeuermann:
### Purpose

Give meaningful error messages and advise instead of having some Segmentation Fault.

### Approach

Check if `dependency[j]-1` is in range of `allVariables` before evaluating `allVariables[dependency[j] - 1])`.

### DIFF
--- a/src/OMSimulatorLib/ComponentFMUCS.cpp
+++ b/src/OMSimulatorLib/ComponentFMUCS.cpp
@@ -237,8 +237,18 @@ oms::Component* oms::ComponentFMUCS::NewComponent(const oms::ComRef& cref, oms::
   component->connectors.push_back(NULL);
   component->element.setConnectors(&component->connectors[0]);
 
-  component->initializeDependencyGraph_initialUnknowns();
-  component->initializeDependencyGraph_outputs();
+  if (oms_status_ok != component->initializeDependencyGraph_initialUnknowns())
+  {
+    logError(std::string(cref) + ": Couldn't initialize dependency graph for initial unknowns.");
+    delete component;
+    return NULL;
+  }
+  if (oms_status_ok != component->initializeDependencyGraph_outputs())
+  {
+    logError(std::string(cref) + ": Couldn't initialize dependency graph for simulation unknowns.");
+    delete component;
+    return NULL;
+  }
 
   // parse modelDescription.xml to get start values before instantiating fmu's
   component->values.parseModelDescription((tempDir / "modelDescription.xml").string().c_str());
@@ -379,6 +389,13 @@ oms_status_enu_t oms::ComponentFMUCS::initializeDependencyGraph_initialUnknowns(
     {
       for (size_t j = startIndex[i]; j < startIndex[i + 1]; j++)
       {
+        if (dependency[j] <= 0 || allVariables.size() < dependency[j])
+        {
+          logError(std::string(getCref()) + ": Dependecies from modelDescription.xml erroneous.");
+          logDebug("Can't find variable for dependency with index " + std::to_string(dependency[j]) + " for  initial unknown " + std::string(initialUnknownsGraph.getNodes()[i]) + "." );
+          logInfo("Use flag --ignoreInitialUnknowns=true to ignore dependencies, but this can cause inflated loop size.");
+          return oms_status_fatal;
+        }
         logDebug(std::string(getCref()) + ": " + getPath() + " initial unknown " + std::string(initialUnknownsGraph.getNodes()[i]) + " depends on " + std::string(allVariables[dependency[j] - 1]));
         initialUnknownsGraph.addEdge(allVariables[dependency[j] - 1].makeConnector(), initialUnknownsGraph.getNodes()[i]);
       }
@@ -423,6 +440,13 @@ oms_status_enu_t oms::ComponentFMUCS::initializeDependencyGraph_outputs()
     {
       for (size_t j = startIndex[i]; j < startIndex[i + 1]; j++)
       {
+        if (dependency[j] <= 0 || allVariables.size() < dependency[j])
+        {
+          logError(std::string(getCref()) + ": Dependecies from modelDescription.xml erroneous.");
+          logDebug("Can't find variable for dependency with index " + std::to_string(dependency[j]) + " for output " + std::string(outputs[i]) + "." );
+          logInfo("Use flag --ignoreInitialUnknowns=true to ignore dependencies, but this can cause inflated loop size.");
+          return oms_status_fatal;
+        }
         logDebug(std::string(getCref()) + ": " + getPath() + " output " + std::string(outputs[i]) + " depends on " + std::string(allVariables[dependency[j] - 1]));
         outputsGraph.addEdge(allVariables[dependency[j] - 1].makeConnector(), outputs[i].makeConnector());
       }
--- a/src/OMSimulatorLib/ComponentFMUME.cpp
+++ b/src/OMSimulatorLib/ComponentFMUME.cpp
@@ -239,8 +239,18 @@ oms::Component* oms::ComponentFMUME::NewComponent(const oms::ComRef& cref, oms::
   component->connectors.push_back(NULL);
   component->element.setConnectors(&component->connectors[0]);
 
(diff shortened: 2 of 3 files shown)


src/OMSimulatorLua/OMSimulatorLua.c
```diff
@@ -892,6 +892,11 @@ static int OMSimulatorLua_oms_addSubModel(lua_State *L)
 
   lua_pushinteger(L, status);
 
+  if (status!=oms_status_ok)
+  {
+    return luaL_error(L, "oms_addSubModel(%s,%s) failed", cref, fmuPath);
+  }
```
lochel, line 898: I guess this should come before pushing the return value `status`.

src/OMSimulatorLib/ComponentFMUCS.cpp
```diff
@@ -379,6 +384,12 @@ oms_status_enu_t oms::ComponentFMUCS::initializeDependencyGraph_initialUnknowns(
     {
       for (size_t j = startIndex[i]; j < startIndex[i + 1]; j++)
       {
+        if (dependency[j] <= 0 || allVariables.size() < dependency[j])
+        {
+          logError("Out of range of vector allVariables.");
+          logInfo("Dependecies from modelDescription.xml could be wrong. Use flag --ignoreInitialUnknowns=true to ignore dependencies.");
+          return oms_status_fatal;
+        }
```
lochel, line 392: What's about `initializeDependencyGraph_outputs`?

AnHeuermann: Yes, I should add this there as well.

src/OMSimulatorLua/OMSimulatorLua.c
AnHeuermann: I don't understand what `lua_pushinteger` is supposed to do.  It pushes the error status to the first position in some Lua stack? So it will save the error and the `luaL_error` displays the error massage and aborts?

lochel: The stack contains all return values, and the returned integer specifies the number of returned values, i.e. size of the stack.

AnHeuermann:
[OpenModelica PR #6803](https://github.com/OpenModelica/OpenModelica/pull/6803) should solve the root cause of the problem, but I'm still in favor of keeping this to handle older FMUs without seg faults.

lochel:
For sure, we will keep the error handling 😄

src/OMSimulatorLib/ComponentFMUCS.cpp
AnHeuermann: Done

src/OMSimulatorLua/OMSimulatorLua.c
AnHeuermann: I moved it up

src/OMSimulatorLib/ComponentFMUCS.cpp
```diff
@@ -237,7 +237,12 @@ oms::Component* oms::ComponentFMUCS::NewComponent(const oms::ComRef& cref, oms::
...
   component->element.setConnectors(&component->connectors[0]);
 
-  component->initializeDependencyGraph_initialUnknowns();
+  if (oms_status_ok != component->initializeDependencyGraph_initialUnknowns())
+  {
+    logError("Couldn't initialize dependency graph for initial unknowns.");
+    delete component;
+    return NULL;
+  }
   component->initializeDependencyGraph_outputs();
```
lochel: Here we should check the return value.

src/OMSimulatorLib/ComponentFMUCS.cpp
```diff
@@ -379,6 +384,12 @@ oms_status_enu_t oms::ComponentFMUCS::initializeDependencyGraph_initialUnknowns(
     {
       for (size_t j = startIndex[i]; j < startIndex[i + 1]; j++)
       {
+        if (dependency[j] <= 0 || allVariables.size() < dependency[j])
+        {
+          logError("Out of range of vector allVariables.");
```
lochel, line 394: Should this rather be a debug message?

src/OMSimulatorLib/ComponentFMUME.cpp
```diff
@@ -239,7 +239,12 @@ oms::Component* oms::ComponentFMUME::NewComponent(const oms::ComRef& cref, oms::
...
   component->element.setConnectors(&component->connectors[0]);
 
-  component->initializeDependencyGraph_initialUnknowns();
+  if (oms_status_ok != component->initializeDependencyGraph_initialUnknowns())
+  {
+    logError("Couldn't initialize dependency graph for initial unknowns.");
+    delete component;
+    return NULL;
+  }
   component->initializeDependencyGraph_outputs();
```
lochel: Here we should check the return value.

src/OMSimulatorLib/ComponentFMUCS.cpp
```diff
@@ -423,6 +434,12 @@ oms_status_enu_t oms::ComponentFMUCS::initializeDependencyGraph_outputs()
...
```
lochel, line 444: ... and this one

src/OMSimulatorLib/ComponentFMUME.cpp
```diff
@@ -393,6 +398,12 @@ oms_status_enu_t oms::ComponentFMUME::initializeDependencyGraph_initialUnknowns(
     {
       for (size_t j = startIndex[i]; j < startIndex[i + 1]; j++)
       {
+        if (dependency[j] <= 0 || allVariables.size() < dependency[j])
+        {
+          logError("Out of range of vector allVariables.");
```
lochel, line 408: ... and this one

src/OMSimulatorLib/ComponentFMUME.cpp
```diff
@@ -437,6 +448,12 @@ oms_status_enu_t oms::ComponentFMUME::initializeDependencyGraph_outputs()
...
```
lochel, line 458: ... and this one

lochel (changes requested):

src/OMSimulatorLib/ComponentFMUCS.cpp
AnHeuermann: If I return an `oms_status_fatal` this should be an error log I guess.

lochel: But then you have to change the error message. The error messages are for the users, not for us :) And there is already the other error message that will be shown:
`Couldn't initialize dependency graph for initial unknowns.`

AnHeuermann:
I updated the error massages. The output looks now like this:

```
debug:   car: resources/0001_car.fmu initial unknown tau depends on gear.phi_support
debug:   car: resources/0001_car.fmu initial unknown tau depends on velSens.flange.f
debug:   car: resources/0001_car.fmu initial unknown tau depends on dragF.rho
debug:   car: resources/0001_car.fmu initial unknown tau depends on dragF.v_small
debug:   car: resources/0001_car.fmu initial unknown tau depends on dragF.startBackward
debug:   car: resources/0001_car.fmu initial unknown tau depends on dragF.startForward
error:   [initializeDependencyGraph_initialUnknowns] car: Dependecies from modelDescription.xml erroneous.
debug:   Can't find variable for dependency with index 116 for initial unknown tau.
info:    Use flag --ignoreInitialUnknowns=true to ignore dependencies, but this can cause inflated loop size.
error:   [NewComponent] car: Couldn't initialize dependency graph for initial unknowns.
debug:   [log level VERBOSE] module FMILIB: Releasing allocated library resources
error:   [oms_RunFile] EVDemo.lua:9: oms_addSubModel(EVDemo.root.car,wbEHVpkg.EV.FirstEV_adapters_car.fmu) failed
debug:   removed temp directory: "/mnt/d/workspace/EMISYS/Demonstratores/EvExample/FirstEV1FMU/temp/EVDemo-g213anm3"
info:    0 warnings
info:    3 errors
```

lochel (approved):
👍

Issue #830, opened 2020-10-13
Endless loop with failing FMI functions


AnHeuermann:
### Description

When simulating a FMU with a loop that can't be solved OMSimulator won't stop the simulation at returned `fmi2Error`.

### Steps to reproduce the behavior

Generate a 2.0 ME FMU from
```modelica
model brokenLoop
  Real x;
  Real y;
equation
  sin(time) = x*x + 0.5*y;
  x*x = 0.5*y;
end brokenLoop;
```
 The loop has solution `y = sin(time)` and `x = sqrt(0.5*sin(time))` which is only solvable for `time<2*pi`. 

```Lua
oms_setCommandLineOption("--deleteTempFiles=true")
oms_setTempDirectory("./temp/")
oms_newModel("model")
oms_addSystem("model.root", oms_system_sc)
-- instantiate FMUs
oms_addSubModel("model.root.system1", "simple.fmu")
-- simulation settings
oms_setResultFile("model", "results.mat")
oms_setStopTime("model", 4)
oms_setFixedStepSize("model.root", 1e-4)
oms_instantiate("model")
oms_initialize("model")
oms_simulate("model")
oms_terminate("model")
oms_delete("model")
```

It will run into an endless loop doing:
```bash
error:   [fmi2logger] system1 (logFmi2Call): fmi2GetReal: terminated by an assertion.
error:   [updateSignals] failed to fetch variable y
assert            | debug   | Solving non-linear system 11 failed at time=3.24830000000243.
|                 | |       | For more information please use -lv LOG_NLS.
error:   [fmi2logger] system1 (logFmi2Call): fmi2CompletedIntegratorStep: terminated by an assertion.
error:   [stepUntil] fmi2_import_completed_integrator_step failed for FMU "model.root.system1"
assert            | debug   | Solving non-linear system 11 failed at time=3.24840000000243.
|                 | |       | For more information please use -lv LOG_NLS.
error:   [fmi2logger] system1 (logFmi2Call): fmi2GetReal: terminated by an assertion.
error:   [updateSignals] failed to fetch variable y
assert            | debug   | Solving non-linear system 11 failed at time=3.24840000000243.
|                 | |       | For more information please use -lv LOG_NLS.
error:   [fmi2logger] system1 (logFmi2Call): fmi2CompletedIntegratorStep: terminated by an assertion.
error:   [stepUntil] fmi2_import_completed_integrator_step failed for FMU "model.root.system1"
```

### Expected behavior

[...]

PR #831, opened 2020-10-13
Stop at error

AnHeuermann:
### Related Issues

#830

### DIFF
--- a/src/OMSimulatorLib/SystemSC.cpp
+++ b/src/OMSimulatorLib/SystemSC.cpp
@@ -623,10 +623,7 @@ oms_status_enu_t oms::SystemSC::stepUntil(double stopTime, void (*cb)(const char
     for (int i=0; i < fmus.size(); ++i)
     {
       fmistatus = fmi2_import_completed_integrator_step(fmus[i]->getFMU(), fmi2_true, &callEventUpdate[i], &terminateSimulation[i]);
-      if (fmi2_status_ok != fmistatus)
-      {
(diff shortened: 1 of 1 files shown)


src/OMSimulatorLib/SystemSC.cpp
```diff
@@ -625,7 +625,11 @@ oms_status_enu_t oms::SystemSC::stepUntil(double stopTime, void (*cb)(const char
     for (int i=0; i < fmus.size(); ++i)
     {
       fmistatus = fmi2_import_completed_integrator_step(fmus[i]->getFMU(), fmi2_true, &callEventUpdate[i], &terminateSimulation[i]);
-      if (fmi2_status_ok != fmistatus) logError_FMUCall("fmi2_import_completed_integrator_step", fmus[i]);
+      if (fmi2_status_ok != fmistatus)
+      {
+        logError_FMUCall("fmi2_import_completed_integrator_step", fmus[i]);
+        return oms_status_error;
+      }
```
lochel, line 642: You could just write `return logError_FMUCall ...` because log_Error returns the error status to simplify those code blocks.

lochel: Do we actually release all resources correctly if exiting `stepUntil` in this way?

AnHeuermann: I bet we don't.

lochel: Is it something we should be worried about?

PR #834, opened 2020-10-19
Silence C compiler warning

AnHeuermann:
### Related Issues

So many warnings when compiling OMSimulator.

### Purpose

Fixed one of these.


adeas31:
I m guessing most of the warnings from FMIL.

lochel:
Which warning did you try to fix?

AnHeuermann:
I would have to rebuild to find the exact error. Basically the compiler is complaining about the space behind the percent symbol:  `% `.

More correct would be to pass the string `""` and repeat it `width - pos + 1`-times.
See https://stackoverflow.com/questions/7899119/what-does-s-mean-in-printf

lochel:
Ok, I don't remember why I put a white space in there. Maybe we should even change it to %c: `printf("%*c %3d%%", width - pos + 1, ']', percent);` What do you think?

PR #836, opened 2020-10-20
Fix and improve some messages

lochel:
### Purpose

Improve error message in case of corrupted dependencies within the modelDescription. This should make it easier to pinpoint the problem in a potentially broken FMU.

I also replaced `oms_status_fatal` with `oms_status_error` since fatal is only used in the FMI interface but not in OMSimulator.

The info message `Use flag --ignoreInitialUnknowns=true to ignore dependencies, but this can cause inflated loop size.` was even displayed in case of broken output dependencies, which was just a copy&past error.


AnHeuermann (approved):
Lgtm

Issue #803, opened 2020-09-03
FMI Cross Check results as HTML for nightly builds


AnHeuermann:
**Is your feature request related to a problem? Please describe.**
For participation in the FMI Cross Check (#739) we wan't to have regular tests with the cross check repo.
To get some history and to save older results we can do something similar we do for OpenModelica already: https://libraries.openmodelica.org/branches/newInst/Modelica_3.2.3/Modelica_3.2.3.html

**Describe the solution you'd like**
A html page could look like this:
![image](https://user-images.githubusercontent.com/38031952/92112572-4727a000-edee-11ea-8627-9edae788fed1.png)

Tinkering with the html we have for OpenModelica until it works would be my suggestion.

lochel:
Looks good! I like it 👍

sjoelund:
Do all of the different models have exactly 42 signals to verify?

AnHeuermann:
Nope, just a placeholder for now. I'm not generating any csv files for singe diffs at the moment and don't count the number of failed variables.
But I'll add this soon

AnHeuermann:
First HTML generated on Jenkins:
https://test.openmodelica.org/jenkins/job/Sandbox/job/Andreas/job/OMSimulator%20FMI%20Cross%20Check/lastSuccessfulBuild/artifact/OMSimulator/testsuite/fmi-cross-check/html/

This took me ages, but it's slowly getting somewhere.

AnHeuermann:
See PRs https://github.com/OpenModelica/OpenModelicaLibraryTesting/pull/4 and https://github.com/OpenModelica/OpenModelicaBuildScripts/pull/6 for more details on the progress.

AnHeuermann:
My latest generated html with #818:
![screencapture-localhost-html-fmi-cross-check-html-2020-09-25-17_22_17](https://user-images.githubusercontent.com/38031952/94285260-bc9d1100-ff53-11ea-8c59-4183d17acfaf.png)

Please note, that on the Jenkins job this file will have less tests passed, since my local machine has more stuff installed than the Docker image used for testing.

AnHeuermann:
Results for Ubuntu Bionic and Focal are now online: https://libraries.openmodelica.org/fmi-crosschecking/OMSimulator/
Windows results are still missing though.

lochel:
Looks good, thanks!

Why are the tests for bionic and focal sorted differently? And why do they use different OpenModelica versions?

[...]

Issue #739, opened 2020-05-19
Participation in FMU Cross-Check


chrbertsch:
It would be highly beneficial if OMSimulator would participate in the FMI Cross Check https://github.com/modelica/fmi-cross-check. For many tools this has significantly improved the maturity of their FMI support.

lochel:
Thanks for reminding us. Actually, it has been on my list for quite some time now. I will try to prioritize it.

lochel:
@chrbertsch Is there some kind of script to support executing all the tests? I guess not since all importing tools are very different.

chrbertsch:
You could use the XC scripts of fmpy as a blueprint:
https://github.com/CATIA-Systems/FMPy/tree/master/fmpy/cross_check

AnHeuermann:
See [ticket 6010](https://trac.openmodelica.org/OpenModelica/ticket/6010) for OpenModelica.
I can probably add OMSimulator there as well with a few additional lines.

AnHeuermann:
I did a quick run with a Python script to see what sort of error messages we get.
I used a Docker Container with Ubuntu 20.04 with only OMSimulator and Python installed.

Basically we have three kind of errors
1. We don't support **FMI 1.0**.
     This is design decision and will not change.
    ```
    error:   [oms_RunFile] oms_extractFMIKind failed
    ```
  
2. JModelica has FMUs which are **missing dynamic libraires**.
    ```
    error:   [fmiLogger] module FMICAPI: Could not load the FMU binary: libgfortran.so.3: cannot open shared object file: No such file or directory
    error:   [instantiate] Could not load "resources/0001_fmu.fmu" which is associated with "model.root.fmu"; it may be corrupted 
or may not support your platform
    error:   [initialize] Model "model" is in wrong model state
    error:   [simulate] Model "model" is in wrong model state
    error:   [terminate] Model "model" is in wrong model state
    error:   [terminate] Model "model" is in wrong model state
    ```
    We can of course install those inside our Docker image for this.

3. We get some **segmentation faults** for SystemModeler and JModelica FMUs or if we reject the FMU:
    ```
    Segmentation fault
    ```
    or
    ```
    error:   [fmiLogger] module FMI2XML: Start attribute is required for this causality, variability and initial combination
    Segmentation fault
    ```
    This needs to be investigated.

4. We reject a lot of FMUs because of **variability of some variables**.
    ```
[...]

PR #839, opened 2020-10-21
Now possible to use wine for testscript

AnHeuermann:
- You can run the Python scripts with "wine /path/to/OMSimulator.exe"
  - Sorted HTML result table

### Related Issues

#739, #803

### Purpose

It's now possible to run the tests with wine from a Linux OS and on Windows.

### Approach

Sorting order: `"Exporting Tool", "Exporting Tool Version", "FMI Version", "FMI Type", "Model Name"`


lochel:
👍

PR #815, opened 2020-09-22
Implement Kinsol as non-linear solver for non-linear loops over components

AnHeuermann:
### Related Issues

#812, #808

### Purpose
A fixed-point iterations is currently the only solver available to solve algebraic loops over components but not suitable for all problems.
Implementing KINSOL as alternative to the fixed-point iteration will enable more FMUs connected to run smoothly.

Issue #808, opened 2020-09-15
Algebraic Loops over FMUs failing


AnHeuermann:
### Description
I have a non-linear algebraic loop with three variables and equations
```modelica
  a   + b   + c    = 0;
  2*a - 3*b + 2*c  = 9;
  a*a + b*b + c*c  = 5;
```
with Solution `a≈0.635425, b≈-1.8, c≈1.16458` or `a≈1.16458, b≈-1.8, c≈0.635425`.
When splitting the loop into two FMUs OMSimulator detects the loop, but can't solve it.

### Steps to reproduce the behavior
I used this mos script to test:
```modelica
loadModel(Modelica);

loadString("model loopOverFMUsA
  Real a, b, c;
  Modelica.Blocks.Interfaces.RealOutput a_out;
  Modelica.Blocks.Interfaces.RealOutput b_out;
  Modelica.Blocks.Interfaces.RealInput c_in;
equation
  c = c_in;
  // Eq. 1 and 2 of loop
  a + b + c=0;
  2*a - 3*b + 2*c=9;
  a_out = a;
  b_out = b;
end loopOverFMUsA;
"); getErrorString();

loadString("model loopOverFMUsB
  Real a, b, c;
  Modelica.Blocks.Interfaces.RealInput a_in;
  Modelica.Blocks.Interfaces.RealInput b_in;
  Modelica.Blocks.Interfaces.RealOutput c_out;
equation
  a = a_in;
  b = b_in;
  // Equation 3 of loop
  a*a + b*b + c*c=5;
  c_out = c;
end loopOverFMUsB;
"); getErrorString();

buildModelFMU(loopOverFMUsA, fmuType="me"); getErrorString();
buildModelFMU(loopOverFMUsB, fmuType="me"); getErrorString();

writeFile("loopsOverFMUs.lua","
oms_setTempDirectory(\"./temp/\")
oms_newModel(\"loopOverFMUs\")
oms_addSystem(\"loopOverFMUs.root\", oms_system_sc)

-- instantiate FMUs
oms_addSubModel(\"loopOverFMUs.root.A\", \"loopOverFMUsA.fmu\")
oms_addSubModel(\"loopOverFMUs.root.B\", \"loopOverFMUsB.fmu\")

-- connections: A -> B
oms_addConnection(\"loopOverFMUs.root.A.a_out\", \"loopOverFMUs.root.B.a_in\")
[...]

PR #848, opened 2020-10-28
Implement Kinsol as solver for non-linear loops over components (#815)

lochel:
### Related Issues

* #808
* #812
* #815
* #847

### Purpose

The changes need to be cleaned up before they can be merged back into master. The functionality seems to be okay though.


AnHeuermann:
Ping

lochel:
This is not flying... #910 will recover the original kinsol implementation from @AnHeuermann and I'll try to find some time for improvements later.

PR #853, opened 2020-10-30
Trying to fix MINGW build with new OMDev

AnHeuermann:
This is now compiling on my machine with mingw. But we don't test the mingw build on Jenkins, but cross-compile for windows.

### DIFF
--- a/src/OMSimulator/CMakeLists.txt
+++ b/src/OMSimulator/CMakeLists.txt
@@ -32,10 +32,6 @@ IF (APPLE)
     $<TARGET_FILE:OMSimulator>)
 ENDIF ()
 
-IF (WIN32 AND MINGW)
-  set(CMAKE_EXE_LINKER_FLAGS "-static -static-libgcc -static-libstdc++")
-ENDIF ()
-
 add_definitions(-DOMS_STATIC)
 target_link_libraries(OMSimulator OMSimulatorLib_static)
 


lochel:
@AnHeuermann What more is needed to actually switch to the new OMDev on Jenkins?

AnHeuermann:
Oh we should be using the new OMDev, only when compiling with gcc we have trouble. But my solution is not good. It will change the size of the executables significantly.

adrpo:
What is the problem with this? I don't get it, it works fine to build OMSimulator on OMDev.

adrpo:
Ah, I remember now, there were some issues with gcc. I will have a look.

AnHeuermann:
We should probably revert [OMTLMSimulator commit 1377e71](https://github.com/OpenModelica/OMTLMSimulator/commit/1377e71f97e73f474904f177f29e7fe7af97b940).
@adrpo What do you think? I was only doing try and error fixes, but the commit ended in the master branch faster then I anticipated.

lochel:
I want to always build OMSimulator with the new OMDev, because it comes with a newer boost version.

AnHeuermann:
It is already building with the new OMDev  using `clang`/`clang++`, or is it not? It's only when you compile it as a standalone version the default command will use `gcc` instead of `clang` when you run it on your Windows PC. And with the newer version of gcc OMSimulator will not build.

So at the moment you have to use
```bash
make config-3rdParty CC=clang CXX=clang++
make config-OMSimulator CC=clang CXX=clang++
make OMSimulator -j4 CC=clang CXX=clang++
```

adrpo:
If I compile with:
`make -f Makefile.omdev.mingw CC=gcc CXX=g++ omsimulator` 
in the top of the OM directory it works fine to compile with gcc.

adrpo:
Hmm, let me do some `git clean -fdx` in OMSimulator and recompile.

adrpo:
Ok, now we're talking:
`C:/OMDev/tools/msys/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lomtlmsimulator`
I guess because we force static linking. I remember fixing this a while back.

adrpo:
#887 should fix the gcc issue.

PR #855, opened 2020-10-30
Test cross-compiled mingw64 and msvc64

lochel:
see #854


AnHeuermann:
Great, I fixed the windows tests by using docker to cross-compile but apparently broke most linux tests when updating them as well :unamused:

At leas when running all but the tlm and python tests locally I only get errors like:
```
-<ssd:Component name="Gain" type="application/x-fmu-sharedlibrary" source="resources/0001_Gain.fmu">
+<ssd:Component name="Gain" type="application/x-fmu-sharedlibrary" source="resources\0001_Gain.fmu">
```
So I need to tell rtest to ignore this kind of differences. But I don't know how rtest is working.

Also there are some permission errors when deleting files on Windows.

lochel (commented):
Should `testsuite/resources/tlm.massSpring/binaries/win64/tlm_massSpring_FMU.libs` be an empty file.

There are 24 such libs in this PR.

sjoelund:
> So I need to tell rtest to ignore this kind of differences. But I don't know how rtest is working.

No, I think that's an error. The SSP standard says:

> For example for an FMU called MyDemoFMU.fmu, that is located in the resources directory of an SSP, the correct URI would be resources/MyDemoFMU.fmu.

And they are supposed to be relative URIs, which means forward slash is the directory separator.

lochel:
Using `generic_string` would require to upgrade to a newer boost version on Windows 😒

PR #857, opened 2020-11-05
Archive artifacts to build.openmodelica.org

AnHeuermann:
Upload artifacts to build.openmodelica.org for Pythons pip installation of OMSimulator python module.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,6 +10,7 @@ pipeline {
     booleanParam(name: 'MSVC64', defaultValue: true, description: 'Build with MSVC64 (often hangs)')
     booleanParam(name: 'MINGW32', defaultValue: false, description: 'Build with MINGW32 (does not link boost)')
     booleanParam(name: 'SUBMODULE_UPDATE', defaultValue: false, description: 'Allow pull request to update submodules (disabled by default due to common user errors)')
+    booleanParam(name: 'UPLOAD_BUILD_OPENMODELICA', defaultValue: false, description: 'Upload install artifacts to build.openmodelica.org/omsimulator')
     string(name: 'RUNTESTS_FLAG', defaultValue: '', description: 'runtests.pl flag')
   }
   stages {
@@ -65,6 +66,7 @@ pipeline {
             sh 'make doc doc-html doc-doxygen'
             sh '(cd install/linux/doc && zip -r "../../../OMSimulator-doc-`git describe --tags --abbrev=7 --match=v*.* --exclude=*-dev | sed \'s/-/.post/\'`.zip" *)'
             archiveArtifacts artifacts: 'OMSimulator-doc*.zip,OMSimulator-linux-amd64-*.tar.gz', fingerprint: true
+            stash name: 'amd64-zip', includes: "OMSimulator-linux-amd64-*.tar.gz"
             stash name: 'docs', includes: "install/linux/doc/**"
           }
         }
@@ -193,6 +195,7 @@ pipeline {
             '''
 
             archiveArtifacts "OMSimulator-linux-i386-*.tar.gz"
+            stash name: 'i389-zip', includes: "OMSimulator-linux-i386-*.tar.gz"
 
             partest()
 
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -465,17 +472,89 @@ EXIT /b 1
...
+            unstash name: 'mingw64-zip'       // includes: "OMSimulator-mingw64*.zip"
+            unstash name: 'win64-zip'         // includes: "OMSimulator-win64*.zip"
+            unstash name: 'osx-zip'           // includes: "OMSimulator-osx*.zip"
+
+            sh "ls *.zip *.tar.gz"
+
+            sshPublisher (
+              publishers: [
+                sshPublisherDesc(
+                  configName: 'OMSimulator-doc',
```
adrpo, line 531: This needs to be "OMSimulator" instead "OMSimulator-doc".

AnHeuermann: Changed it.

AnHeuermann:
Now the artifacts will only be uploaded if the Jenkins build get's started with `UPLOAD_BUILD_OPENMODELICA`. We should only upload "good" builds, e.g. tagged versions or if a new feature is available that the Python version needs.

sjoelund:
You can check in Jenkins if you are building a pull request, a branch, or a tag and only upload automatically based on those.

AnHeuermann:
> You can check in Jenkins if you are building a pull request, a branch, or a tag and only upload automatically based on those.

Yeah. But I don't want to upload artifacts for all successful builds on master. To many files with only a few changes. The lastest installs will always be available as Jenkins artifacts. But we probably want to have more then only the tagged commits.
At least this is my opinion.

sjoelund:
> Yeah. But I don't want to upload artifacts for all successful builds on master. To many files with only a few changes. The lastest installs will always be available as Jenkins artifacts. But we probably want to have more then only the tagged commits.

You can filter out master :)

adrpo (approved):
OK by me!

lochel (changes requested):
If I get it right, the final download link would be something like this:

    https://build.openmodelica.org/omsimulator/linux/i386/OMSimulator-linux-amd64-v2.0.0-242-ga90fcb4.tar.gz

I would like to change the naming and directory structure to something like this:

    https://build.openmodelica.org/omsimulator/linux-i386/OMSimulator-2.0.0.post245.tar.gz

sjoelund:
> I would like to change the naming and directory structure to something like

Please follow semver.org for names so they can be sorted properly. Dots after 2.0.0 are a no-no. Never put a name like `post245`, but rather use `post.245`.

lochel:
> Please follow semver.org for names so they can be sorted properly. Dots after 2.0.0 are a no-no. Never put a name like `post245`, but rather use `post.245`.

No, we don't use that version schema. You can open an issue to discuss changing the version schema if you want. But please don't clutter this PR.

AnHeuermann:
Changing the directories to `omsimulator/linux-i386/` (and analogous for the others) is no problem. I'll add a commit for that.

Changing `OMSimulator-linux-amd64-v2.0.0-242-ga90fcb4.tar.gz` has all sort of implications. We should have only one format for OMSimulator version (something like `OMSimulator v2.0.0.post236-ga34ab0e-linux`), the name of the archive we save on Jenkins for the latest install and the installs we want to save on build.openmodelica.org.

So what is the format we wan't to have for this? Whatever it is we need to change it at some additional places (e.g. FMI Cross Check on Jenkins).

But I would say let's only make the first change in this PR and discuss how the name for version and artifacts should look like an change that there.

Jenkinsfile
```diff
@@ -465,17 +473,90 @@ EXIT /b 1
...
+                      sourceFiles: 'OMSimulator-mingw32*.zip'),
+                    sshTransfer(
+                      remoteDirectory: "win-mingw64/",
+                      sourceFiles: 'OMSimulator-mingw64*.zip'),
+                    sshTransfer(
+                      remoteDirectory: "osx/",
+                      sourceFiles: 'OMSimulator-osx*.zip'),
+                    sshTransfer(
+                      remoteDirectory: "win-msvc64/",
+                      sourceFiles: 'OMSimulator-win64*.zip')
```
lochel, line 555: Do we need to create the directories on build.openmodelica.org manually?

sjoelund: Not if you create the directory locally and transfer the `sourceFiles: 'linux-i386/OMSimulator-linux-i386-*.tar.gz'` I think. Also note that it's not `i368`

AnHeuermann: The ssh transfer creates all directories. At least it did when I uploaded to the wrong location.

lochel (approved):

PR #866, opened 2020-11-11
Update README.md

lochel:


README.md
```diff
@@ -4,7 +4,17 @@ The OMSimulator project is a FMI-based co-simulation environment that supports o
 
 ## Downloads
 
-Pre-compiled binaries are available for [Windows, Linux, and Mac](https://test.openmodelica.org/jenkins/job/OMSimulator/job/master/lastSuccessfulBuild/artifact/).
+### Full package
+
+Pre-compiled binaries are available for [Windows, Linux, and Mac](https://build.openmodelica.org/omsimulator/).
```
AnHeuermann, line 9: I would prefer to have both available, the pre-compiled binaries for specific versions/tags and the latest artifacts from Jenkins.

If we only wan't to have the binaries from build.openmodelica.org we should remove the [archive artifacts step](https://github.com/OpenModelica/OMSimulator/blob/master/Jenkinsfile#L68) from the Jenkinsfile.

AnHeuermann: :+1:

lochel: I re-added a pointer to the latest build.

PR #887, opened 2020-11-20
fix gcc build on new OMDev, link with the tlm lib dll

adrpo:
### Purpose

Fix gcc build on new OMDev

### Approach

Link OMSimulator with the full dll path so that you can do it either statically or dynamically.


adrpo:
`mingw-clang -j4` on a slower computer is as fast as `mingw-gcc -j16` on the fast computer :)

lochel (approved):

lochel:
Does the cross-compilation still use the old OMDev?

adrpo:
> Does the cross-compilation still use the old OMDev?

Cross compilation uses Linux, I have no idea what gcc they are using. Let me check.

adrpo:
It seems is GCC 7.3.0: '-- The C compiler identification is GNU 7.3.0. In the new OMDev we have GCC 10.2.0 and Clang 11.0.0.

lochel:
I see! Is there any reason to keep the cross-compiled version?

adrpo:
For the cross compilation no boost is used it seems: `-- Using C++17 instead of Boost`.
Also the cross compilation on Linux is done in 36 seconds vs 3 minutes on Windows.

adrpo:
I think the cross compilation is way faster, more stable and we have more Linux machines and only 4 machines.

lochel:
Don't we use C++17 on Windows?

adrpo:
I think we should, let me check.

lochel:
`std::filesystem::path::generic_string` is part of C++17 but the cross-compiled mingw version fails with the following error (see #888):
```
CMakeFiles/OMSimulatorLib.dir/objects.a(ComponentFMUCS.cpp.obj):ComponentFMUCS.cpp:(.text+0x6f3e): undefined reference to `std::experimental::filesystem::v1::__cxx11::path::generic_string() const'
```

Your new windows targets succeed for both gcc and clang.

adrpo:
The only 'boost' I can find here (mingw-gcc):
https://test.openmodelica.org/jenkins/blue/rest/organizations/jenkins/pipelines/OMSimulator/branches/PR-887/runs/5/nodes/47/steps/258/log/?start=0
is: `-- Using C++17 instead of Boost`.

For mingw-clang seems boost is used which seems strange:
https://test.openmodelica.org/jenkins/blue/rest/organizations/jenkins/pipelines/OMSimulator/branches/PR-887/runs/5/nodes/48/steps/252/log/?start=0

adrpo:
I guess one could update the gcc cross compiler in the mingw-cross image to fix that issue you're having in #888.

lochel:
Or we could just run it on Windows 😃

adrpo:
That works too. Some jobs might have to wait for the Windows machine to be available but it should work fine.

lochel:
@sjoelund do you want to update the mingw-cross image?

adrpo:
Ah, it seems that the cross compilation doesn't build the TLM lib.

lochel:
You are right. We must change that and include tlm in the mingw builds.

sjoelund:
If you want to cross-compile tlm you will need to rewrite it's makefiles to use a configure script or cmake. It has a nonfunctional detection of architecture, etc as far as I remember.

As for Windows machines. We used to build OMS on that, but it failed more than half the time, right?

I suppose the mingw image could be updated. But don't need you need a lot of ifdef/etc to compile on things like CentOS6 anyway in that case since all OSes won't be updated? You should be able to resolve that by just replacing \ with / on Windows.

lochel:
I think OMSimulator used to fail for MSVC but that seem to be fixed for quite some time now.

CentOS6 etc. works fine with #888.

So it it will be easiest to just switch back from cross-compilation to a Windows machine for the mingw builds.

CMakeLists.txt
```diff
@@ -181,10 +181,16 @@ ELSE() # use it from the install folder where is properly installed for MacOS
 ENDIF()
 set(OMTLMSIMULATOR_3RDPARTY_INCLUDEDIR ${PROJECT_SOURCE_DIR}/OMTLMSimulator/3rdParty/misc/include)
 IF(WIN32)
-  set(OMTLM_LINKFLAGS omtlmsimulator wsock32 ws2_32)
-ELSE()
+  IF (MINGW)
+    # adrpo: link with the full path to the dll on MINGW
```
AnHeuermann, line 185: That was exactly what I did but couldn't get it into the CMakeLists how I wanted it. :+1:

AnHeuermann (approved):
This is compiling fine for me

adrpo:
Just let me know how you want to handle this so I know if I need to remove some stages.

lochel:
I would like to replace `mingw64-cross` with either `mingw64-clang` or `mingw64-gcc`. I guess we can keep the other one as optional. And can we also compile a mingw32 version with the new OMDev?

adrpo:
Hopefully OK now. Please check and let me know @lochel @AnHeuermann.

adrpo:
Can you spot where the problem is?! I don't' see it right now. I'll have another look tomorrow.

AnHeuermann:
I couldn't see any errors inside Blue Ocean, but in the [classic view of Pipeline Steps](https://test.openmodelica.org/jenkins/job/OMSimulator/job/PR-887/7/flowGraphTable/) I found a [Windows batch script that failed](https://test.openmodelica.org/jenkins/job/OMSimulator/job/PR-887/7/execution/node/447/):

```bash
hudson.AbortException: script returned exit code 1
	at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.handleExit(DurableTaskStep.java:658)
	at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.check(DurableTaskStep.java:604)
	at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.run(DurableTaskStep.java:548)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

This is inside mingw64-gcc branch, but that shouldn't run, or not?

AnHeuermann:
What git version is installed on the Windows machines?
We try to execute
```cmd
C:\OMDev\tools\msys\usr\bin\sh --login -i -c 'cd ${env.WORKSPACE}/install/mingw && zip -r "../../OMSimulator-mingw64-`git describe --tags --abbrev=7 --match=v*.* --exclude=*-dev | sed 's/-/.post/'`.zip" *'
```
and 
`git describe --tags --abbrev=7 --match=v*.* --exclude=*-dev | sed 's/-/.post/'`.zip" *'` is not working on older git versions.

lochel:
@AnHeuermann @adrpo Any idea how to fix it?

AnHeuermann:
The batch script
```bat
C:\OMDev\tools\msys\usr\bin\sh --login -i -c 'cd ${env.WORKSPACE}/install/mingw && zip -r "../../OMSimulator-mingw64-`git describe --tags --abbrev=7 --match=v*.* --exclude=*-dev | sed 's/-/.post/'`.zip" *'
```
is failing. I guess we need to update the `git` on the windows machines.

lochel:
Okay, can you do it? Or don't you have access to it?

AnHeuermann:
On machine w3 it is
```
$ git --version
git version 2.29.2.windows.2
```
which is up to date. So that's not the problem.

I now found the [error message from Jenkins](https://test.openmodelica.org/jenkins/job/OMSimulator/job/PR-887/8/execution/node/456/log/):
{{{
c:\dev\jenkins\ws\OMSimulator_PR-887>C:\OMDev\tools\msys\usr\bin\sh --login -i -c 'cd ${env.WORKSPACE}/install/mingw   && zip -r "../../OMSimulator-mingw64-`git describe --tags --abbrev=7 --match=v*.* --exclude=*-dev | sed 's/-/.post/'`.zip" *' 
sh: cannot set terminal process group (-1): Inappropriate ioctl for device
sh: no job control in this shell
sh: ${env.WORKSPACE}/install/mingw: bad substitution
}}}

So maybe some Jenkins specific error? I have no idea what this means

AnHeuermann:
We could also try to split `void buildOMS()` (see [Code](https://github.com/OpenModelica/OMSimulator/pull/887/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R712)) into two functions. I had a problem once whit a function that was using `sh` in one if branch and `bat` in another one. But I got a different error message there and Jenkins was not accepting my Jenkinsfile at all.

lochel:
Great, thanks @adrpo! Can it be merged? Or do you want to make more changes?

adrpo:
You can merge it now. It seems a bit unstable but we'll see.

Issue #878, opened 2020-11-18
Use forward slashes in Uri on Windows


AnHeuermann:
### Description

The URI used in ssp on Windows is not correct. It contains backslashes in stead of slashes.

Check for example the output of the failing test `build / mingw64-cross / test / deleteStartValues.lua – OMSimulator` in https://test.openmodelica.org/jenkins/blue/organizations/jenkins/OMSimulator/detail/PR-855/17/tests:

```
-<ssd:Component name="Gain" type="application/x-fmu-sharedlibrary" source="resources/0001_Gain.fmu">
+<ssd:Component name="Gain" type="application/x-fmu-sharedlibrary" source="resources\0001_Gain.fmu">
```

### Steps to reproduce the behavior

In OMSimulator/testsuite/OMSimulator/deleteStartValues.lua change `-- mingw: no` to `-- mingw: yes` and run it with partests.pl

```bash
   cd testsuite/partest/
   ./runtests.pl -j4
```

### Expected behavior

According to the ssp standard:

> For example for an FMU called MyDemoFMU.fmu, that is located in the resources directory of an SSP, the correct URI would be resources/MyDemoFMU.fmu.

### Version and OS

- Version: [OMSimulator v2.0.0.post258-g3135d7a-mingw]
- OS: [Win10, 64bit with OMDev mingw]

lochel:
This has been addressed in #888.

PR #888, opened 2020-11-21
Test boost::path::generic_string

lochel:


AnHeuermann (approved):
We should merge this PR. I tested one example from the testsuite and it will fix the problem described in https://github.com/OpenModelica/OMSimulator/issues/878.

PR #896, opened 2020-12-17
Activating all tests on MINGW

AnHeuermann:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/879

### Purpose

Activating test to see our progress on the windows tests.

PR #899, opened 2020-12-18
Fixing error introduced in 6aace93

AnHeuermann:
Fixxing error from 6aace93.
  - That's why we need to test on Windows again

Jenkins test from #896 
reported
```
Stacktrace
Output mismatch (see stdout for details)
Standard Output
+ buses.py                                                                          ... execution failed
==== Log /tmp/oms-rtest-adrpo33/apitmp_1398/log-buses.py
Traceback (most recent call last):
  File "buses.py", line 9, in <module>
    oms = OMSimulator()
  File "C:/dev/jenkins/ws/OMSimulator_PR-896/install/mingw/lib/OMSimulator/OMSimulator.py", line 17, in __init__
    close(ddlDir)
NameError: name 'close' is not defined
== 1 out of 1 tests failed [api, time: 1]
```

Of course it has to be `ddlDir.close()`.

### DIFF
--- a/src/OMSimulatorPython/OMSimulator.py
+++ b/src/OMSimulatorPython/OMSimulator.py
@@ -10,11 +10,11 @@ class OMSimulator:
       dirname = os.path.dirname(__file__)
       omslib = os.path.join(dirname, "@OMSIMULATORLIB_DIR_STRING@", "@OMSIMULATORLIB_STRING@")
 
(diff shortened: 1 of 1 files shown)


AnHeuermann:
This time I tested it on my Windows machine with a test that is not skipped and therefore always returning true 🤦

Issue #735, opened 2020-05-13
Add Parameter-Mapping support to Parameter-Bindings in SSP


arun3688:
### Purpose
At present, We support inline Parameter binding in SSP, both at the top System Level and Component level, But without parameter mapping, This needs to be implemented, as with parameter mapping a user can set the top Level parameters to different components in the subSystem. This needs to be supported both inline and as well as through external file (.ssm)

robha67:
Could such a mapping of parameters also be done directly to a Modelica model in OpenModelica and not only an FMU? I realize that this might be more of an OpenModelica feature but could the OMSimulator support for SSP somehow be used to establish this link?

lochel:
This is still relevant.

PR #904, opened 2020-12-30
import parameter mapping from .ssm and inline

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/735

### Purpose

This PR supports importing` parameter mapping` from `.ssm` file and `inline`, according to `SSP-1.0`

### Approach 
Currently only importing of `parameter mapping` is supported.

### example
```
<ssd:ParameterBindings>
    <ssd:ParameterBinding source="resources/import_parameter_mapping.ssv">
        <ssd:ParameterMapping source="resources/import_parameter_mapping.ssm" />
    </ssd:ParameterBinding>
</ssd:ParameterBindings>

```

### DIFF
--- a/src/OMSimulatorLib/System.cpp
+++ b/src/OMSimulatorLib/System.cpp
@@ -528,8 +528,21 @@ oms_status_enu_t oms::System::importFromSSD(const pugi::xml_node& node, const st
         }
         else
         {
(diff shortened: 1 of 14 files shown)


arun3688:
@robha67  This PR implements  parameter mapping support in OMSimulator, i have tested your test cases and the results are looking correct, please verify the results

Issue #865, opened 2020-11-11
Option to select which variables to export (.csv, .mat)


markaren:
**Is your feature request related to a problem? Please describe.**
Some systems may contain a large number of variables, but the user may only care about a few.
Writing all variables (say many thousands), when the user only needs a handful is wasteful and impacts performance. It also makes it more difficult for the user to find the variables to plot after the fact.

**Describe the solution you'd like**
Make it possible to supply a configuration file (e.g. XML to make it consistent with SSP. Could also be a SSP annotation) with the variables to export. 

**Additional context**
Vico([example](https://github.com/NTNU-IHB/Vico/blob/master/examples/gunnerus/trajectory/LogConfig.xml)) and libcosim([example](https://github.com/open-simulation-platform/libcosim/blob/master/tests/data/LogConfig.xml)) allows this, with the additional option to specify how often variables from an individual model should be exported (e.g. each 10 time-step).
In these systems, no file means export everything, a file with no simulators/components means log nothing.

lochel:
There is already the option to filter signals for the result file. This can in certain caises drastically improve performance. 
* [setSignalFilter](https://openmodelica.org/doc/OMSimulator/master/html/OMSimulatorPython.html#setsignalfilter)
* [addSignalsToResults](https://openmodelica.org/doc/OMSimulator/master/html/OMSimulatorPython.html#addsignalstoresults)
* [removeSignalsFromResults](https://openmodelica.org/doc/OMSimulator/master/html/OMSimulatorPython.html#removesignalsfromresults)

It is also possible to subsample signals to reduce the number of data point for a simulation:
* [setLoggingInterval](https://openmodelica.org/doc/OMSimulator/master/html/OMSimulatorPython.html#setlogginginterval)

We could also implement a xml-based filter as described above.

lochel:
@robha67 This could also be of interest to you.

markaren:
Thanks! Could you provide an example of a signalFilter string that would ignore everything?

lochel:
An empty string should do it: `oms.setSignalFilter(<model>, "")`

markaren:
I'm using the SSP interface :/

lochel:
Of course you do 😄 

```xml
	<ssd:DefaultExperiment startTime="0.000000" stopTime="1.000000">
		<ssd:Annotations>
			<ssc:Annotation type="org.openmodelica">
[...]

PR #915, opened 2021-01-25
export signal filters to xml file

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/865

### Purpose

This PR exports signal filters to an external xml file `"signalFilter.xml"`  in `ssp` which is stored in the resources directory and referenced in `ssd` file. If no signal Filters are used the default signal filter is set to all `".*"`

### Approach
 `oms_setSignalFilter()`  deprecated 
`oms_addSignalsToResults()` use to add signals to results
`oms_removeSignalsFromResults()`  use to remove signals from result files

### Example
**signalFilter.xml**
```
<?xml version="1.0" encoding="UTF-8"?>
<oms:SignalFilter version="1.0">
	<oms:Variable name="model.root.System2.Input_1" />
	<oms:Variable name="model.root.Gain.u" />
	<oms:Variable name="model.root.Gain.k" />
</oms:SignalFilter>
```

**ssdFile**
```
<oms:Annotations>
    <oms:SimulationInformation resultFile="model_res.mat" loggingInterval="0.000000" bufferSize="10" signalFilter="resources/signalFilter.xml" />
</oms:Annotations>
```

### TODO

- [x] add signal causality to signal filter
- [x] add signal kind to signal filter
- [x] recover `oms_setSignalFilter` to not break older scripts
- [x] Fix `oms_importSnapshot` (see `test03.[py,lua]` and `snapshort.[py,lua]`)


src/OMSimulatorLib/ComponentTable.h
```diff
@@ -94,6 +96,7 @@ namespace oms
     std::unordered_map<unsigned int /*result file var ID*/, unsigned int /*allVariables ID*/> resultFileMapping;
     double time;
     double storedTime;
+    std::vector<ComRef> filteredSignals;  ///< filtered signals set via oms_addSignalsToResults()
```
lochel, line 99: Isn't this sort of the same as `std::unordered_map<ComRef, bool> exportSeries;`?

arun3688: @lochel  this should be fixed with this https://github.com/OpenModelica/OMSimulator/pull/915/commits/ae0fab74593c4754886d1fa6bf51e92013f1f19b

testsuite/OMSimulator/import_export_signalFilter.lua
```diff
@@ -0,0 +1,201 @@
...
+-- 					</ssc:Annotation>
+-- 				</ssd:Annotations>
+-- 			</ssd:DefaultExperiment>
+-- 		</ssd:SystemStructureDescription>
+-- 	</oms:ssd_file>
+-- 	<oms:signalFilter_file name="resources/signalFilter.xml">
+-- 		<oms:SignalFilter version="1.0">
+-- 			<oms:Variable name="model.root.System2.Input_1" />
+-- 			<oms:Variable name="model.root.Gain.u" />
+-- 			<oms:Variable name="model.root.Gain.k" />
```
lochel, line 188: Would it make sense to include the signal type as well?

testsuite/OMSimulator/import_export_signalFilter.lua
```diff
@@ -0,0 +1,201 @@
...
+oms_initialize("model")
+readFile("model_res.csv")
+
+oms_terminate("model")
+oms_delete("model")
+
+-- Result:
+-- <?xml version="1.0"?>
+-- <oms:snapshot>
+-- 	<oms:ssd_file name="SystemStructure.ssd">
```
lochel, line 73: We should simplify the snapshot by replacing `oms:ssd_file` etc. with just `oms:file` for all files.

lochel (changes requested):
And the exportSnapshot api should be extended to allow to query the files without content, e.g. returning
```
<?xml version="1.0"?>
<oms:snapshot>
  <oms:file name="SystemStructure.ssd" />
  <oms:file name="resources/signalFilter.xml" />
</oms:snapshot>
```

And it should be extended to be able to return just a specific file. Sort of like the `list()` command which could be short hand for `exportSnapshot(<model>, "SystemStructure.ssd", true)`.

testsuite/OMSimulator/import_export_signalFilter.lua
arun3688: this is fixed with https://github.com/OpenModelica/OMSimulator/pull/915/commits/0b50246e2a14b3ddeef743c41a03cfbcae5075d9

testsuite/OMSimulator/snapshot.py
```diff
@@ -106,6 +106,15 @@
 ## 			</ssv:Parameters>
 ## 		</ssv:ParameterSet>
 ## 	</oms:ssv_file>
+## 	<oms:signalFilter_file name="resources/signalFilter.xml">
```
lochel, line 109: Please change it to `<oms:file name="resources/signalFilter.xml">`

testsuite/OMSimulator/snapshot.py
```diff
@@ -106,6 +106,15 @@
...
+## 		<oms:SignalFilter version="1.0">
+## 			<oms:Variable name="snapshot.root.add.u1" type="real" kind="input" />
+## 			<oms:Variable name="snapshot.root.add.u2" type="real" kind="input" />
+## 			<oms:Variable name="snapshot.root.add.y" type="real" kind="output" />
+## 			<oms:Variable name="snapshot.root.add.k1" type="real" kind="parameter" />
+## 			<oms:Variable name="snapshot.root.add.k2" type="real" kind="parameter" />
```
lochel, line 115: SSP and FMI use `Real` instead for `real`, don't they?

testsuite/api/test_omsExport.py
```diff
@@ -57,6 +57,7 @@ def printStatus(status, expected):
 ## Result:
 ## status:  [correct] ok
 ## status:  [correct] ok
+## warning: [oms_setSignalFilter] is deprecated , use [oms_addSignalsToResults] to add signals and [oms_removeSignalsFromResults] to remove signals
```
lochel, line 60: remove the white-space before the comma

lochel (changes requested):
just some minor issues

testsuite/api/test_omsExport.py
arun3688: done with this https://github.com/OpenModelica/OMSimulator/pull/915/commits/ae8174960b62f59c22f92e87c336328bd0194111

testsuite/OMSimulator/snapshot.py
arun3688: fixed with this https://github.com/OpenModelica/OMSimulator/pull/915/commits/ae8174960b62f59c22f92e87c336328bd0194111

testsuite/OMSimulator/snapshot.py
arun3688: we can fix this in another PR

src/OMSimulatorLib/Model.cpp
```diff
@@ -721,14 +721,30 @@ oms_status_enu_t oms::Model::importSignalFilter(const std::string& filename)
...
+  {
+    // load from memory
+    std::vector<oms::ComRef> filteredSignals;
+    system->getFilteredSignals(filteredSignals);
+    removeSignalsFromResults(".*"); // disable all signals
+    for (auto const &signal : filteredSignals)
+    {
+      addSignalsToResults(signal.c_str());
+    }
+  }
```
lochel, line 747: This is not doing anything. `system->getFilteredSignals()` only returns the signals that are already filtered.

lochel:
OMSimulator/snapshot.[py,lua] and api/test03.lua are failing because the snapshot api doesn't include the signal filter xml file. That is why we introduced the new `exportSnapshot` `importSnapshot` api.

lochel:
I think we should restore the setSignalFilter function just to not break older scripts. It should call
> oms_removeSignalsFromResults(model, ".*")
> addSignalsToResults(model, regex);

lochel (changes requested):
The changes look good now. Just a few minor things to do before we can merge it:

### TODO

- [ ] 
add signal causality to signal filter
- [ ] add signal kind to signal filter
- [ ] recover `oms_setSignalFilter` to not break older scripts


- [ ] Fix `oms_importSnapshot` (see `test03.[py,lua]` and `snapshort.[py,lua]`)

src/OMSimulatorLib/Model.cpp
arun3688: yes but i think it is needed, when we do not export the `snapshot` to `ssp` and the user set a signal filter, then we need to load it from memory,  otherwise all signals will be exported to signal filter,  we do it for `.ssv file ` loading it from memory, should'nt we do the same for signal filter as well

```
oms_removeSignalsFromResults("snapshot", ".*")
oms_addSignalsToResults("snapshot", "snapshot.root.add.u1")
snapshot = oms_exportSnapshot("snapshot")
oms_setReal("snapshot.root.add.u1", 3.5)
oms_importSnapshot("snapshot", snapshot)

```

src/OMSimulatorLib/Model.cpp
```diff
@@ -143,6 +144,19 @@ oms_status_enu_t oms::Model::rename(const ComRef& cref, const ComRef& newCref)
...
+oms::Variable* oms::Model::getVariable(const ComRef& cref) const
+{
+  // renaming any subcomponent
+  oms::ComRef tail(cref);
+  oms::ComRef front = tail.pop_front();
+
+  if (system->getCref() == front)
+    return system->getVariable(tail);
+
+  logWarning("failed to find signal \"" + std::string(cref) + "\"");
```
lochel, line 156: This is just for debugging and should be removed before merging the pull request.

src/OMSimulatorLib/Model.cpp
```diff
@@ -632,8 +646,15 @@ void oms::Model::exportSignalFilter(pugi::xml_node &node) const
   {
     pugi::xml_node oms_variable = node.append_child(oms::ssp::Version1_0::oms_Variable);
     oms_variable.append_attribute("name") = signal.c_str();
-    //oms_variable.append_attribute("type") = getTypeString(connector->getType()).c_str();
-    //oms_variable.append_attribute("kind") = getCausalityString(connector->getCausality()).c_str();
+
+    oms::ComRef cref(signal);
+    cref.pop_front();
```
lochel, line 651: The signal names within the signal filter shouldn't start with the model name. So this should be considered as a workaround and fixed before merging the pull request.

src/OMSimulatorLib/Model.h
```diff
@@ -159,7 +162,7 @@ namespace oms
     std::string resultFilename; ///< default <name>_res.mat
     Clock clock;
 
-    std::string signalFilter = ".*"; ///< default
+    std::string signalFilterFileName = "resources/signalFilter.xml";
 
     pugi::xml_node snapShot; ///< top level snapshot node which contains ssd, ssv and ssm as child nodes
```
lochel, line 167: Why do we have this xml node in here?

adeas31:
What is the status of this PR? @lochel mentioned that it is needed to continue the interactive simulation server work. Can we prioritize this PR. See also https://github.com/OpenModelica/OpenModelica/pull/7237

lochel:
@adeas31 We continued the development in a new PR: #985 and I will probably merge it today 😃

Issue #909, opened 2021-01-22
Jupyter Notebook kernel crashing with OMSimulator


AnHeuermann:
### Description

When running OMSimulator within a Jupyter Notebook the kernel will die on Ubuntu Focal while it works perfectly fine on Windows 10 with Anaconda.

### Steps to reproduce the behavior

I created a dockerfile to reproduce the problem in an "empty" Ubuntu Focal installation
```dockerfile
FROM ubuntu:focal

ARG DEBIAN_FRONTEND=noninteractive
ENV DISPLAY=host.docker.internal:0.0

RUN groupadd -g 1000 nonRootUser \
     && useradd -m -u 1000 -g nonRootUser nonRootUser

RUN apt-get update \
    && apt-get install python3 python3-pip -qy\
    && apt-get install firefox -qy

USER nonRootUser
ENV PATH="$PATH:/home/nonRootUser/.local/bin"

RUN pip3 install requests \
    && pip3 install notebook \
    && pip3 install OMSimulator

WORKDIR /home/nonRootUser
CMD jupyter notebook
```

```bash
docker build --tag=omsimulator-jupyter .
docker run --rm -it omsimulator-jupyter
```

Create a new Notebook and add a Cell with
```python
from OMSimulator import OMSimulator
oms = OMSimulator()
```
If you run this cell the kernel will die.

The output from console
```bash
andreas@DESKTOP-OSJIVA0:~/workspace/Testitesttest$ docker run --rm -it omsimulator-jupyter
[I 10:27:23.014 NotebookApp] Writing notebook server cookie secret to /home/nonRootUser/.local/share/jupyter/runtime/notebook_cookie_secret
[I 10:27:23.155 NotebookApp] Serving notebooks from local directory: /home/nonRootUser
[I 10:27:23.155 NotebookApp] Jupyter Notebook 6.2.0 is running at:
[I 10:27:23.155 NotebookApp] http://localhost:8888/?token=938a1c14529b24bfbb6cfedff5da77229c7b2932e13363e4
[I 10:27:23.155 NotebookApp]  or http://127.0.0.1:8888/?token=938a1c14529b24bfbb6cfedff5da77229c7b2932e13363e4
[I 10:27:23.155 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 10:27:23.161 NotebookApp] 
    
    To access the notebook, open this file in a browser:
[...]

PR #916, opened 2021-01-26
Don't add static C / C++ lib to dynamic libs

AnHeuermann:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/909

### Purpose

Stop Python from crashing when loading libOMSimulator.so on Linux.

### Approach

Don't link the static C and C++ libs to the dynamic build target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,6 @@ ELSEIF(APPLE)
   set(PLATFORM_STRING "mac")
 ELSE()
   set(PLATFORM_STRING "linux")
-  set(STATIC_LIBGCC "-static-libgcc")
 ENDIF()
 
 # Don't allow in-source build
--- a/src/OMSimulatorLib/CMakeLists.txt
+++ b/src/OMSimulatorLib/CMakeLists.txt
@@ -154,7 +154,7 @@ IF (WIN32 AND MSVC)
   add_definitions(-DBOOST_ALL_DYN_LINK)
 ENDIF ()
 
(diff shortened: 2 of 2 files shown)


AnHeuermann:
We seriously need to sort out the problems inside the testsuite. There are tests failing at random. Every time a restart is something different, but totally unrelated to my changes.

Maybe they are a sign that there is still something wrong going on with the way we compile OMSimulator. I mean the error with Python was in OMSimulator for nearly a year and non of our Python tests managed to catch it.

lochel:
I haven't seen this error before:
```
OMSimulator: /var/lib/jenkins2/ws/OMSimulator_PR-916/src/OMSimulatorLib/MatVer4.cpp:95: void oms::appendMatVer4Matrix(FILE*, long int, const char*, size_t, size_t, const void*, oms::MatVer4Type_t): Assertion `header.type == (isBigEndian() ? 1000 : 0) + type' failed.
Aborted (core dumped)
```

AnHeuermann:
Now it's `build / linux64-asan / test / import_parameter_mapping_inline.lua – OMSimulator` test that fails...

AnHeuermann:
I don't see why not linking with `-static-libstdc++ -static-libgcc  -lstdc++fs` causes the linux-asan tests to fail.
On test run 5  `import_parameter_mapping_inline.lua` and `import_parameter_mapping_from_ssm.lua` are failing, on run 6 (no changes, just restarted the job) it is only `tlm1d.lua` and on run 7 (again a restart) `import_parameter_mapping_inline.lua` is failing, again.

@sjoelund You did the initial commit where you added those flags to OMSimulator. Do you have some ideas?

sjoelund:
> @sjoelund You did the initial commit where you added those flags to OMSimulator. Do you have some ideas?

Probably to make the shared object work on systems that do not have the older libstdc++ installed. But you're right that perhaps this would interfere with applications using a different C++ library. This is why you create C libraries instead :)

As for why ASAN fails... I don't know. It's still running build+test in the same docker image.

lochel:
@AnHeuermann on run 8 all tests passed - still without any changes compared to the previous runs. This isn't looking good... 😕

AnHeuermann:
We should merge it for now and fix the pip installation on Linux. And then put this with a high priority on the todo list.

Issue #917, opened 2021-01-27
Random test failures in PRs and on the master


AnHeuermann:
### Description

We seem to have some tests in our testsuite that are failing at random in our PRs.
I looked through some of my last PRs and found the following tests that are failing and passing on different runs of Jenkins without any changes to the code bases:

PR-916, commit e6f58b8
- `build / linux64-asan / test / import_parameter_mapping_inline.lua – OMSimulator`
- `build / mingw64-gcc / test / import_parameter_mapping_from_ssm.lua – OMSimulator`
- `build / linux64-asan / test / tlm1d.lua – tlm`

Also we fail some tests on the master branch, e.g.
master, commit 2f00f94
- `build / linux64-asan / test / import_parameter_mapping_inline.lua – OMSimulator`
- `build / mingw64-gcc / test / import_parameter_mapping_inline.lua – OMSimulator`

master, commit e814c71
- `build / mingw32-gcc / test / exportSSMTemplate.py – OMSimulator`

master, commit 6616083
- `build / mingw64-gcc / test / Enumeration.lua – OMSimulator`

### Expected behavior

When a PR passes the test it should also have a successful build on the master.

AnHeuermann:
And of course I can't reproduce the exact error with asan:
```
$docker run --rm -it -v /home/andreas/workspace/OMSimulatorStandalone:/OMSimulator docker.openmodelica.org/build-deps:v1.13 bash

#export ASAN=ON
#echo $ASAN
#cd /OMSimulator/
#make config-3rdParty -j
#make config-OMSimulator -j ASAN=ON
#make OMSimulator -j DISABLE_RUN_OMSIMULATOR_VERSION=1
#./install/linux/bin/OMSimulator --version
#exit

$docker run --rm -it -v /home/andreas/workspace/OMSimulatorStandalone:/OMSimulator --cap-add SYS_PTRACE --privileged --oom-kill-disable -m 1024m --memory-swap 1024m docker.openmodelica.org/build-deps:v1.13 bash

#export ASAN=ON
#echo $ASAN
#cd /OMSimulator/
#./install/linux/bin/OMSimulator --version
#make -C testsuite difftool resources
#cd testsuite/partest
#./runtests.pl -asan -nocolour -with-xml -j16
```

Only `tlm3d.lua` is failing (or to slow).

But if I add `ulimit -v 6291456` I start to get problems. Of course I don't have 6GB memory per processor available, but could it be that the limit on memory per processor is a problem?

[...]

PR #918, opened 2021-01-27
Check if malloc / calloc returned NULL

AnHeuermann:
### Related Issues

#917

### Purpose

See if we still get random errors in the testsuite.

### DIFF
--- a/src/OMSimulatorLib/MatVer4.cpp
+++ b/src/OMSimulatorLib/MatVer4.cpp
@@ -119,6 +119,8 @@ oms::MatVer4Matrix* oms::readMatVer4Matrix(FILE* file)
   MatVer4Type_t type = (MatVer4Type_t) (matrix->header.type % 100);
   size_t size = sizeofMatVer4Type(type);
   matrix->data = malloc(matrix->header.mrows * matrix->header.ncols * size);
+  if (!matrix->data)
+    return NULL;
   fread(matrix->data, size, matrix->header.mrows*matrix->header.ncols, file);
 
   return matrix;
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -699,7 +699,7 @@ def numPhysicalCPU() {
 }
 
 void partest(cache=true, extraArgs='') {
-  echo "cache: ${cache}, asan: ${env.ASAN}"
+  echo "cache: ${cache}, asan: ${env.ASAN}, running on node: ${env.NODE_NAME}"
   sh """
   make -C testsuite difftool resources
   cp -f "${env.RUNTESTDB}/"* testsuite/ || true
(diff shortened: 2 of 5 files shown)


AnHeuermann:
It's quiet hard to test something that is only failing every now and then. Did this improve anything or is it still the same?

src/OMSimulatorLib/SystemSC.cpp
```diff
@@ -204,6 +204,7 @@ oms_status_enu_t oms::SystemSC::instantiate()
       n_states += nStates.back();
       nEventIndicators.push_back(fmus.back()->getNumberOfEventIndicators());
 
+      /* TODO: Will this memory ever get freed??? */
```
lochel, line 207: Yes, it is done here:

https://github.com/OpenModelica/OMSimulator/blob/94de8fcd4b5dfd53bae9e6f7e4c2925bb017a347/src/OMSimulatorLib/SystemSC.cpp#L396-L403

src/OMSimulatorLib/MatVer4.cpp
```diff
@@ -119,6 +122,11 @@ oms::MatVer4Matrix* oms::readMatVer4Matrix(FILE* file)
   MatVer4Type_t type = (MatVer4Type_t) (matrix->header.type % 100);
   size_t size = sizeofMatVer4Type(type);
   matrix->data = malloc(matrix->header.mrows * matrix->header.ncols * size);
+  if (!matrix->data)
+  {
+    printf("Out of memory");
+    return NULL;
+  }
```
lochel, line 129: This shouldn't print any error, but the function calling it should check for the return value and issue a warning. But I guess this is just for testing right now.

AnHeuermann: But `matrix->data` is used in the next line to read data into it. If we are out of memory this should be a problem.

lochel: Yes, you should return, but not print from inside the function.

AnHeuermann: Ah, I see.

lochel (approved):
LGTM :+1:

lochel:
Why does the jenkins report show the asan test being skipped?

AnHeuermann:
I think it is a bug in Blueocean. If I go to the classic pipeline view in Jenkins I can see that the build step of `linux-asan` is evaluated and successful.

lochel:
Ok, then just go ahead and merge this pull request.

AnHeuermann:
We will see if this solves anything. At least the asan test should not be able to run out of memory now. It's using only one thread and has all unlimited virtual memory . If our tests keep failing we can see on which node they are running and at least rule out a memory problem at this place.

PR #922, opened 2021-01-28
Fix FMI Cross Check with Wine

AnHeuermann:
### Related Issues

All tests from fmi-cross-check were broken for the tagged version without a commit hash in the OMSimulator version.
See https://test.openmodelica.org/jenkins/job/Sandbox/job/Andreas/job/OMSimulator-FMI-Cross-Check/

Also the Windows test with wine is failing, see #924.

### Approach

- Just add a "unknown" commit into the html.
- Link with `-static-libstdc++` on MINGW
 - Add check to Jenkins to see if wine can execute MINGW OMSimulator binary

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -574,6 +574,25 @@ EXIT /b 1
       }
     }
 
+    stage('cross-compilation') {
+      agent {
+        docker {
+          image 'anheuermann/ompython:wine-bionic'
+          label 'linux'
+          alwaysPull true
+        }
+      }
+      environment {
+        HOME = "${env.WORKSPACE}"
+      }
+      steps {
+        unstash name: 'mingw64-install'
+        sh """
+        wine64 install/mingw/bin/OMSimulator.exe --version
+        """
+      }
+    }
+
     stage('upload') {
       parallel {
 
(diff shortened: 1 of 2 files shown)


AnHeuermann:
@sjoelund Do you happen to know how to set up the Jenkins stage at https://github.com/OpenModelica/OMSimulator/pull/922/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R577-R593 so wine will create a configure directory?

```bash
export HOME=/var/lib/jenkins1
wine64 install/mingw/bin/OMSimualtor.exe --version
wine: '/var/lib/jenkins1' is not owned by you, refusing to create a configuration directory there
script returned exit code 1
```

adrpo:
You seem to have a typo: `OMSimualtor.exe` -> `OMSimulator.exe` but maybe that's not the problem.

AnHeuermann:
Well, the typo was definitely not helping.
I hope this run is successful, otherwise I will just use the docker image that is used to run the FMI Cross Check with wine: `anheuermann/ompython:wine-bionic`. That should work at least... I hope... Maybe....

sjoelund:
I think the reason we use things like `export HOME="$PWD"` in other jobs is that `env.WORKSPACE` is the path on the machine that controls the job. Not the path used on the machine running the job. If you want something simple, I guess do `HOME=/tmp`

lochel (approved):
LGTM

PR #923, opened 2021-01-28
remove the stable/ target altogether - temporary

lochel:
- additionally copy stable releases to
  https://build.openmodelica.org/omsimulator/stable/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -680,7 +680,7 @@ EXIT /b 1
 
         stage('upload-artifacts-stable') {
           when {
-            expression { return shouldWeUploadArtifacts() && params.STABLE}
+            expression {return params.STABLE}
             beforeAgent true
           }
           agent {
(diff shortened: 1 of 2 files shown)


lochel:
@AnHeuermann maybe it's even better to always upload to the nightly/ directory and additionally to stable/ in case of releases. Might be less confusing and would make it easier for the pip package to find artifacts.

AnHeuermann:
It's a lot of duplicate code, but I don't see how we can simplify that.

Jenkinsfile
```diff
@@ -678,6 +678,58 @@ EXIT /b 1
           }
         }
 
+        stage('upload-artifacts-stable') {
```
AnHeuermann, line 681: Stage  `upload-artifacts` will also upload to stable/

lochel: Not with the changes from this pr, right?

AnHeuermann:
Also we don't need to upload the same files twice. Can't we just create a link from nightly/ to stable (or the other way around)?

lochel:
> Also we don't need to upload the same files twice. Can't we just create a link from nightly/ to stable (or the other way around)?

I guess so, yes. However, it just doesn't matter since we have so few releases and they are very small in size.

Jenkinsfile
AnHeuermann: Oh yes. If `params.STABLE` is true this stage will rune and upload to `stable/`.
The stage above will use `DEPLOYMENT_PREFIX = getDeploymentPrefix()` and get `nightly/`

AnHeuermann:
Maybe I should check how big the folder at https://libraries.openmodelica.org/fmi-crosschecking/ has grown. Maybe we need some housekeeping as well.

AnHeuermann:
> Or we just remove the stable/ target altogether

That's fine with me.

Issue #919, opened 2021-01-27
System input connector can only be connected to one component input connector


lochel:
### Description

The screenshot below shows an error message when connecting a system input connector can to a second component input connector.

> [addConnection] Connector is already connected: u_s

### Expected behavior

It should be possible to connect a system input connector to several sub-model input connectors.

### Screenshots

![image](https://user-images.githubusercontent.com/8457302/106047670-2e44c180-60e4-11eb-8b27-50e86010d30d.png)

### Version and OS

- Connected to OMCompiler v1.17.0-dev.359+ge7fca47128
- Connected to OMSimulator v2.0.0.post286-g94de8fc-linux
- WSL2 - Ubuntu 20.04

lochel:
This is fixed :hurray:

PR #925, opened 2021-01-28
allow input connectors to multiple components connector

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/919

### Purpose

This PR allows` input connector` to be connected with `multiple component input connector` which has no previous connections defined.  According to `SSP-1.0 ` specification multiple connections to already connected  connectors of type input/output/parameter should not exist.

### DIFF
--- a/testsuite/api/connections.lua
+++ b/testsuite/api/connections.lua
@@ -53,7 +53,7 @@ printStatus(status, 0)
 status = oms_addConnection("model.wc.sc1.u1", "model.wc.sc2.y1")
 printStatus(status, 0)
 
---Connecting input to input (illegal)
+--Connecting output to output (illegal)
 status = oms_addConnection("model.wc.sc1.y", "model.wc.sc2.y3")
 printStatus(status, 3)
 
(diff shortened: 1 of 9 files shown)


lochel (commented):
Haven't read the code yet, but I can confirm that it seems to work:
image

Just the error message in case of connecting output connectors several times needs to be improved. I propse something like:
> [addConnection] Connector y is already connected to addSat.y

testsuite/api/connections.lua
```diff
@@ -90,8 +90,7 @@ printStatus(status, 0)
 -- status:  [correct] error
 -- error:   [addConnection] Type mismatch in connection: sc1.u2 -> sc2.y2
 -- status:  [correct] error
--- error:   [addConnection] Connector is already connected: sc1.u1
--- status:  [correct] error
+-- status:  [wrong] ok
```
lochel, line 93: @arun3688 This looks wrong.

testsuite/api/connections.py
```diff
@@ -90,8 +90,7 @@ def printStatus(status, expected):
 ## status:  [correct] error
 ## error:   [addConnection] Type mismatch in connection: sc1.u2 -> sc2.y2
 ## status:  [correct] error
-## error:   [addConnection] Connector is already connected: sc1.u1
-## status:  [correct] error
+## status:  [wrong] ok
```
lochel, line 93: @arun3688 and this as well - but should be the same issue as the one above.

arun3688: @lochel this should be fixed with this https://github.com/OpenModelica/OMSimulator/pull/925/commits/d646dbc5ac9c9c687298960bcdd2617225d7d9e6

testsuite/api/connections.lua
arun3688: @lochel this should be fixed with this d646dbc

lochel (approved):
Nice! 🚀

lochel:
The error messages are relevant and helpful:

image

PR #931, opened 2021-02-04
Interactive simulation server: OMSimulatorServer.py

lochel:
### Purpose

This server application is intended for interactive simulations. It will simulate a SSP file and takes commands via zmq connection to control the simulation.

### TO DO

- [x] Implement `ZMQ_REP` socket using zeroMQ
- [x] Implement `ZMQ_PUB` socket using zeroMQ
- [x] Send the simulation progress data.


lochel:
@adeas31 It seems that the doStep api doesn't quite work. The simulation doesn't progress... I will fix that tomorrow. Besides that, the server should work fine.

PR #932, opened 2021-02-04
New api: oms_doStep

lochel:
### Related Issues

* Required for #931 

### Purpose

There is no api support to execute a single (macro) step.

### TO DO

- [x] C-API
- [x] Python binding
- [x] wc system
- [x] sc system
- [x] tlm system
- [x] documentation


lochel:
@AnHeuermann @arun3688 I know; this pull request is very painful to review - but it was also very painful to create 😉

lochel:
I see... no one wants to have a look at this new amazing api :stuck_out_tongue_winking_eye:

adrpo:
You have a lot of constants like 1e16 in the code. Should those be hardcoded?

lochel:
Thanks @adrpo! The 1e16 is only used for micro adjusting the final simulation step. Maybe it shouldn't be hard-coded - but I'll keep it for now.

PR #940, opened 2021-02-10
Modify loadSnapshot so that it can be used to rename models

lochel:
### Related Issues

@adeas31 is there a related issue? Couldn't find it...

### Purpose

Extend the scope of oms_loadSnapshot. It was not possible to use it to rename models. This should be fixed with this pull request.

### Approach

1. Check the name in the snapshot
2. Check if it is different to to current name
3. If so: check if the new model exist already in the scope
4. rename/load snapshot


lochel:
@adeas31 Please adapt OMEdit to the changes. `oms_loadSnapshot` has a new signature and returns the new model name since it might actually change now.

arun3688 (approved):

PR #943, opened 2021-02-11
Refactor loadSnapshot

lochel:
### Purpose

This requires the xml snapshot only once to be parsed.
It also provides a cleaner separation between the Scope and Model class.

@arun3688 I changed the version check from a warning to an error. But I think we had it intentionally as warning. Do you remember it?


arun3688:
@lochel as far i remember `fmigo` had a different draft versions in their `SSP's` `eg. Draft20170606` see. https://github.com/fmi-tools/FMIGo/blob/master/tools/ssp/examples/parameters-inline/SystemStructure.ssd.

arun3688 (approved):

lochel:
I think we shouldn't try to import version which do do not support.

PR #944, opened 2021-02-11
Install OMSimulatorServer.py to build/share/OMSimulator/scripts

adeas31:
### Related Issues

* OpenModelica/OpenModelica#7135

### Purpose

Allow OMEdit to use the python script file after installation.

### Approach

Copy the script file to share folder.


lochel:
Thanks, I'll have to test if it interferes with the stand-alone version of OMSimulator.

lochel (approved):
I changed it so that the server is only installed within the OpenModelica super project.

PR #948, opened 2021-02-15
Simplify ssv related code in class Model

lochel:
This is just simplifying some of the existing code.


src/OMSimulatorLib/System.cpp
```diff
@@ -1689,6 +1692,9 @@ bool oms::System::copyResources()
 
 void oms::System::getAllResources(std::vector<std::string>& resources) const
 {
+  if (!values.empty() && !Flags::ExportParametersInline())
```
arun3688, line 1695: this will not work, and values.empty() will always be false and the ssv files will not be pushed to resources

arun3688: the ssv file can be pushed directly in `exportoFile()` in `Model.cpp` which is easily understandable i guess

src/OMSimulatorLib/System.cpp
```diff
@@ -1689,6 +1692,9 @@ bool oms::System::copyResources()
...
+    resources.push_back("resources/" + std::string(this->getFullCref()) + ".ssv");
```
arun3688, line 1696: the ssv file can be pushed directly in `exportoFile()` in `Model.cpp` which is easily understandable i guess

src/OMSimulatorLib/System.cpp
lochel: Why will values.empty() always be false?

src/OMSimulatorLib/System.cpp
lochel: values contain all the values that will be exported to the ssv file, right? Therefore, it would make sense to move everything ssv related to that class. I am not quite sure why it didn't work, but I would still like to move those things there.

PR #950, opened 2021-02-16
Adding python helper for FMI Cross Check result filtering

AnHeuermann:
### Related Issues
Small helper I used to create the PR to https://github.com/modelica/fmi-cross-check for https://github.com/OpenModelica/OMSimulator/issues/739.

Not important, but nice to have for the next release.

### Purpose

Reduces size of result files that are bigger then 1 MB by iteratively halving the size of the result file until the size is smaller then 1 MB


lochel (approved):
Fine with me, but you should add a docstring to the file. Otherwise I will soon be wondering what this is and ping you 😆

PR #955, opened 2021-02-17
Try to fix installation within the OpenModelica super project

lochel:
The installation of the simulation server doesn't work on Windows. It works fine on Linux though.

This is an attempt to fix it. The simulation server should only be installed if it is compiled as part of the OpenModelica super project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,12 +85,12 @@ ifeq ($(OMBUILDDIR),)
 else
 	TOP_INSTALL_DIR=$(OMBUILDDIR)
 	CMAKE_INSTALL_PREFIX=-DCMAKE_INSTALL_PREFIX=$(OMBUILDDIR)
-ifeq ($(host_short),)
-	HOST_SHORT=
-else
-	HOST_SHORT_OMC=$(host_short)/omc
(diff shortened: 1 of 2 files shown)


adeas31 (approved):

lochel:
This doesn't work as expected. It will also install the shared files if compiled "stand-alone". @OpenModelica/omsimulator-developers Does anyone know how we can check in cmake reliably if we are compiling from within the OpenModelica super project or not?

sjoelund:
> Does anyone know how we can check in cmake reliably if we are compiling from within the OpenModelica super project or not?

Create a cmake variable that is by default false and set it from OpenModelica when calling `cmake`?

lochel:
> Create a cmake variable that is by default false and set it from OpenModelica when calling `cmake`?

There must be a better way. I am pretty sure that this information is already somehow available.

adrpo:
Check OMBUILDDIR?

adeas31:
@lochel can you test if this works in stand-alone case?

lochel:
@adeas31 it works fine in the stand-alone version 👍

adeas31:
Good. This works from the OpenModelica super project as well and copies the server file.

PR #961, opened 2021-02-22
create oms:snapshot from ssp

arun3688:
### Purpose

The purpose of this PR is to create a `oms:snapshot` internally from `SSP` package, which allows to load all the resources like `.ssv, .ssm and signalFilter.xml ` directly from oms:snapsshot


src/OMSimulatorLib/Scope.cpp
```diff
@@ -224,7 +233,56 @@ oms_status_enu_t oms::Scope::importModel(const std::string& filename, char** _cr
...
+        pugi::xml_document signalFilter_doc;
+        pugi::xml_parse_result result = signalFilter_doc.load_file(entry.path().c_str());
+        if (!result)
+          return logError("loading \"" + entry.path().filename().generic_string() + "\" to <oms:snapshot> failed (" + std::string(result.description()) + ")");
+
+        pugi::xml_node signalfilter_node = signalFilter_doc.document_element(); // oms:SignalFilter
+        signalFilter_file.append_copy(signalfilter_node);
+      }
+    }
+  }
```
lochel, line 281: This should be simplified. Seems like all cases do exactly the same.

src/OMSimulatorLib/Model.h
```diff
@@ -75,7 +75,7 @@ namespace oms
     oms_status_enu_t exportSnapshot(const ComRef& cref, char** contents);
     oms_status_enu_t exportSSVTemplate(const ComRef& cref, const std::string& filename);
     oms_status_enu_t exportSSMTemplate(const ComRef& cref, const std::string& filename);
-    oms_status_enu_t importFromSSD(const pugi::xml_node& node);
+    oms_status_enu_t importFromSnapshot(const pugi::xml_node& oms_snapshot);
```
lochel, line 78: Maybe we shouldn't use the snapshot node, but a `unordered_map<filename, xml_node>`. @arun3688 what do you think?

lochel (changes requested):

src/OMSimulatorLib/Scope.cpp
arun3688: yes this can be simplified

src/OMSimulatorLib/Model.h
arun3688: We can use `unordered_map` , but I like the current `<oms_snapshot>`, Is it because of the performance issue ? or some other advantages?

lochel:
@adrpo looks promising, thanks for your help!

src/OMSimulatorLib/Model.h
arun3688: this is done in this commit https://github.com/OpenModelica/OMSimulator/pull/961/commits/b307b78706008dba6e31fbeb34ea8fb7407fd0a3

src/OMSimulatorLib/Scope.cpp
arun3688: this is done with this https://github.com/OpenModelica/OMSimulator/pull/961/commits/b307b78706008dba6e31fbeb34ea8fb7407fd0a3

lochel:
I removed some of my latest commits. Now it should work fine again and we will keep it as a backup for #965.

lochel (approved):
We decided to merge it as it is and continue with a few other smaller pull requests to get it in good shape.

PR #966, opened 2021-02-25
Encapsulate the snapshot implementation

lochel:
### Related Issues

#961 and #965

### Purpose

Encapsulating the snapshot object with a new class and clean interface.


src/OMSimulatorLib/Model.cpp
```diff
@@ -197,9 +198,14 @@ oms_status_enu_t oms::Model::importSnapshot(const char* snapshot)
...
   pugi::xml_node oms_snapshot = doc.document_element(); // oms:snapshot
 
-  std::unordered_map<std::string, pugi::xml_node> snapshotFiles;
+  Snapshot newSnapshot;
+  // create the snapshot interface
   for (const auto& it : oms_snapshot.children())
-    snapshotFiles[it.attribute("name").as_string()] = it;
+  {
+    newSnapshot.importResourcesXML(it.attribute("name").as_string(), it);
+  }
```
lochel, line 206: All this should be moved to `Snapshot::import(const char* snapshot)`

arun3688 (approved):

PR #968, opened 2021-02-25
Fix issues with suffixes in oms::ComRef

lochel:


lochel:
The `operator+` is still a bit problematic. `ComRef("model.root:start") + ComRef("fmuA")` will return `ComRef("model.root:start.fmuA")` which obviously isn't very good.

Issue #945, opened 2021-02-12
Import/export of snapshots


lochel:
### Purpose

For proper text view/editing in OMEdit an extension of the snapshot functionality is required. It is necessary to handle complete and partial snapshots. It is also necessary to export/import descriptions of systems and components.

### import snapshot

New function signature:
- [x] `oms_status_enu_t oms_importSnapshot(const char* ex_cref, const char* snapshot);`

The import function can be used to import a complete snapshot (this is already implemented):
- [x] `> status = oms_importSnapshot("model", snapshot);`

The import function can be used to import a partial snapshot, i.e. a given file:
- [ ] `> status = oms_importSnapshot("model:resources/signalFilter.xml", snapshot);`
- [ ] `> status = oms_importSnapshot("model.system.component:SystemStructure.ssd", snapshot);`

### export snapshot
This function is returning the snapshot of the given model or just a partial snapshot, i.e. a specific resource file.

New function signature:
- [x] `oms_status_enu_t oms_exportSnapshot(const char* ex_cref, char** snapshot);`

It can be used to query a specific file, e.g. resources/signalFilter.xml:
- [x] `> status = oms_exportSnapshot("model:resources/signalFilter.xml", &snapshot);`

It can be used to query a complete snapshot (without the actual resources, e.g. FMUs) (this is already implemented):
- [x] `> status = oms_exportSnapshot("model", &snapshot);`

### ~list resources~

~This new function will return a xml file containing all the resources filename, e.g. ssd files, ssv files, signal filter. It will not contain the components themselves, e.g. no FMU paths.~
 
- `oms_status_enu_t listResources(const char* cref, char** contents);`

arun3688:
this issue is fixed with https://github.com/OpenModelica/OMSimulator/pull/977

PR #971, opened 2021-03-01
export oms_snapshot using snapshot interface

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/945

### Purpose

The purpose of this PR is to export `<oms_snapshot>` using the `snapshot interface class`


lochel (approved):
This needs more work obviously, because the nodes still get created at the root level and the snapshot object isn't really used. But that might have to wait until the next pull request.

PR #985, opened 2021-03-10
export signalFilter to ssp

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/865

### Purpose

This PR exports signal filters to an external xml file `"signalFilter.xml"`  in `ssp` which is stored in the resources directory and referenced in `ssd` file. If no signal Filters are used the default signal filter is set to all `".*"`

### Example
**signalFilter.xml**
```
<?xml version="1.0" encoding="UTF-8"?>
<oms:SignalFilter version="1.0">
	<oms:Variable name="model.root.Gain.u" type="Real" kind="input" />
	<oms:Variable name="model.root.Gain.k" type="Real" kind="parameter" />
	<oms:Variable name="model.root.System2.Input_1" type="Real" kind="input" />
</oms:SignalFilter>
```


src/OMSimulatorLib/Connector.cpp
```diff
@@ -288,3 +289,12 @@ bool oms::operator!=(const oms::Connector& v1, const oms::Connector& v2)
 {
   return !(v1 == v2);
 }
+
+oms::Variable* oms::Connector::getVariable()
+{
+  oms_signal_type_enu_t type = this->getType();
+  oms_causality_enu_t kind = this->getCausality();
+  oms::Variable * var = new oms::Variable(type, kind);
```
lochel, line 297: This memory is never freed.

arun3688:
@lochel , This should be merged

src/OMSimulatorLib/Model.cpp
```diff
@@ -378,6 +379,11 @@ oms_status_enu_t oms::Model::exportSnapshot(const oms::ComRef& cref, char** cont
     updateParameterBindingsToSSD(system_node, ssvNode, true);
     // TODO ssm file
   }
+  else
+  {
+    // delete ssv node from <oms:snapshot> if flag not set
+    snapshot.deleteResourceNode("resources/" + std::string(this->getCref()) + ".ssv");
+  }
```
lochel, line 386: Why do we generate the node in the first place?

src/OMSimulatorLib/Model.cpp
```diff
@@ -1211,3 +1211,34 @@ oms_status_enu_t oms::Model::removeSignalsFromResults(const char* regex)
...
+  if (!system)
+    return;
+
+  std::vector<oms::Connector> filteredSignals;
+  system->getFilteredSignals(filteredSignals);
+
+  for (auto const& signal : filteredSignals)
+  {
+    pugi::xml_node oms_variable = node.append_child(oms::ssp::Version1_0::oms_Variable);
+    oms_variable.append_attribute("name") = signal.getFullName().c_str();
```
lochel, line 1226: Should we pop the model name?

src/OMSimulatorLib/Model.cpp
```diff
@@ -1211,3 +1211,34 @@ oms_status_enu_t oms::Model::removeSignalsFromResults(const char* regex)
...
+
+  removeSignalsFromResults(".*"); // disable all signals
+  for (pugi::xml_node_iterator it = oms_signalfilter.begin(); it != oms_signalfilter.end(); ++it)
+  {
+    if (std::string(it->name()) == oms::ssp::Version1_0::oms_Variable)
+      addSignalsToResults(it->attribute("name").as_string());
+  }
+
+  return oms_status_ok;
+}
```
lochel, line 1244: We should add all signals in case that the signal filter file doesn't exist.

testsuite/api/test03.py
```diff
@@ -100,13 +100,14 @@ def printStatus(status, expected):
...
 ## 				<oms:Annotations>
-## 					<oms:SimulationInformation resultFile="test_res.mat" loggingInterval="0.000000" bufferSize="10" signalFilter=".*" />
+## 					<oms:SimulationInformation resultFile="test_res.mat" loggingInterval="0.000000" bufferSize="10" signalFilter="resources/signalFilter.xml" />
 ## 				</oms:Annotations>
 ## 			</ssc:Annotation>
 ## 		</ssd:Annotations>
 ## 	</ssd:DefaultExperiment>
 ## </ssd:SystemStructureDescription>
 ##
+## error:   [getResourceNode] Failed to find node "resources/signalFilter.xml"
```
lochel, line 110: Is this because of the old api `loadSnapshot`?

testsuite/OMSimulator/importStartValues.lua
```diff
@@ -156,6 +155,28 @@ oms_delete("importStartValues")
 --       </ssv:Parameters>
 --     </ssv:ParameterSet>
 --   </oms:file>
+--   <oms:file
+--     name="resources/signalFilter.xml">
+--     <oms:SignalFilter
+--       version="1.0">
+--       <oms:Variable
+--         name="importStartValues.root.C1"
```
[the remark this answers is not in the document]
arun3688: @lochel i am ok with full cref as well, but if you want you can remove it

testsuite/api/test03.py
arun3688: yes this is because of `loadSnapshot()` API, it should be replace with `importSnapshot()`

src/OMSimulatorLib/Model.cpp
arun3688: we should handle cases for` signalFilterFilename= ".*" `

src/OMSimulatorLib/Model.cpp
arun3688: this is because of the current signature of `exportToSSD(pugi::xml_node& node, pugi::xml_node& ssvNode, Snapshot& snapshot)`, And this should be changed when we restructure the `exportToFile` to use `snapshot` interface

src/OMSimulatorLib/Model.cpp
lochel: After discussion with Arun: I will revert commit 8f73aba

PR #996, opened 2021-03-25
Fix initial unknowns

lochel:
### Related Issues

Not reported.

### Purpose

The initial unknowns were flagged as wrong for certain FMUs. This was apparently wrong and caused by a lookup issue with value references which are not unique in FMI.

### Approach

Do the loopup with name instead of vr, because the name is unique according to the FMI specification.


arun3688:
@robha67 This PR fixes the initial unknown problem in SAAB test case. Thanks lennart

PR #998, opened 2021-03-25
Fix the initial unknowns once and for all

lochel:
### Related Issues

Not reported; or already closed.

### Purpose

* Detect wrong initial unknowns if possible and report them.
* Only consider `--ignoreInitialUnknowns=true` for FMUs with bad initial unknowns.
* Re-enable AircraftVehicleDemonstrator


lochel:
@AnHeuermann Thanks for your support! I think this is now in good shape and could be merged soon.

PR #1001, opened 2021-03-30
fix renaming connectors and values

arun3688:
### Related Issues

https://github.com/OpenModelica/OpenModelica/pull/7274
https://github.com/OpenModelica/OpenModelica/issues/7111

### Purpose

This PR, fixes the exporting of snapshot when renaming `system, subsystem or submodules`  which will be used by OMEdit text editing of ssp files

### Approach
When renaming `system or subsytem or submodule` , the connector owners are not updated which caused the exportsnapshot to fail, Also the values in ssv files are also updated when renaming.

### DIFF
--- a/src/OMSimulatorLib/Scope.cpp
+++ b/src/OMSimulatorLib/Scope.cpp
@@ -435,24 +435,14 @@ oms_status_enu_t oms::Scope::importSnapshot(const oms::ComRef& cref, const char*
   if (newCref)
     *newCref = NULL;
 
-  oms::ComRef tail(cref);
-  oms::ComRef front = tail.pop_front();
-
-  oms::ComRef modelCref(front);
-  modelCref.pop_suffix();
-
-  oms::Model* model = oms::Scope::GetInstance().getModel(modelCref);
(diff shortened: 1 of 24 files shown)


src/OMSimulatorLib/Scope.cpp
```diff
@@ -435,24 +435,14 @@ oms_status_enu_t oms::Scope::importSnapshot(const oms::ComRef& cref, const char*
...
   if (!model)
-    return logError_ModelNotInScope(front);
+    return logError_ModelNotInScope(cref);
 
-  oms_status_enu_t status;
-  if (tail.isEmpty() && front.hasSuffix())
-    status = model->importSnapshot(oms::ComRef(":" + front.suffix()), snapshot);
-  else
-    status = model->importSnapshot(tail, snapshot);
+  oms_status_enu_t status = model->importSnapshot(snapshot);
```
lochel, line 442: This should probably return the new cref.

arun3688: @this is done with this https://github.com/OpenModelica/OMSimulator/pull/1001/commits/b84662d4cbc823fde943a044052c69b8a9ef5b70

PR #1002, opened 2021-04-01
Added saab Embrace use case to test suite

robha67:
### Related Issues

Not related to any explicit isssue

### Purpose

Common/sharable test-case thought to capture industry needs on the OMSimulator. 

Saab use-case to be included in the test suite. The included SSP contains linux FMUs, parameter data in a SSV file, and a SSM file providing the parameter bindings. The use case is executed using the included python script. The simulation boundary conditions are expressed in the included csv file.


lochel:
Thanks @robha67!

PR #1006, opened 2021-04-15
allow renaming model cref in snapshot

arun3688:
### Purpose

This PR allows renaming `model cref` in `snapshot` which should support renaming model cref in text editing in `OMEdit`.

### DIFF
--- a/src/OMSimulatorLib/Model.cpp
+++ b/src/OMSimulatorLib/Model.cpp
@@ -187,7 +187,7 @@ oms_status_enu_t oms::Model::loadSnapshot(const pugi::xml_node& node)
   return oms_status_ok;
 }
 
-oms_status_enu_t oms::Model::importSnapshot(const oms::ComRef& cref, const char* snapshot_, char** newCref)
+oms_status_enu_t oms::Model::importSnapshot(const char* snapshot_, char** newCref)
 {
   if (!validState(oms_modelState_virgin))
     return logError_ModelInWrongState(getCref());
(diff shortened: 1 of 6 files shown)


src/OMSimulatorLib/Model.cpp
```diff
@@ -208,9 +208,6 @@ oms_status_enu_t oms::Model::importSnapshot(const char* snapshot_, char** newCre
   ComRef new_cref = ComRef(ssdNode.attribute("name").as_string());
   std::string ssdVersion = ssdNode.attribute("version").as_string();
 
-  if (new_cref != getCref())
```
lochel, line 211: Check here if `new_cref` is already in scope or not.

arun3688: @this is fixed with https://github.com/OpenModelica/OMSimulator/pull/1006/commits/174b6d09a23fd39d49d4b5ea57ad9dc64fdbc982

src/OMSimulatorLib/Model.cpp
```diff
@@ -216,8 +216,9 @@ oms_status_enu_t oms::Model::importSnapshot(const char* snapshot_, char** newCre
...
-    return logError("this API cannot be used to rename a model");
+  // check new_cref exists in scope
+  if (new_cref != getCref() && Scope::GetInstance().getModel(new_cref))
+    return logError("renaming model name \"" + std::string(new_cref) + "\" failed as it already exists in the scope");
```
lochel, line 221: return logError("Renaming the model \"" + std::string(getCref()) + "\" to \"" + std::string(new_cref) + "\" failed because another model with the same name already exists in the scope.");

arun3688: fixed with https://github.com/OpenModelica/OMSimulator/pull/1006/commits/37e754c71c844976f3f1cceed2d1ec20cf15d383

Issue #1008, opened 2021-04-17
Renaming SSP crashes OMEdit


max-privato:
### Steps to reproduce the behavior
- Load test1
- right-click on  Gru_Test3 and choose rename
- try to rename into Gru
- => OMEdit crashes

### Version and OS
Tested on 1,18,0-dev 224 64 bit 
Win 10

- Version: 
- OS: [e.g. Win10, 64bit]

Test file:
[Test1.zip](https://github.com/OpenModelica/OMSimulator/files/6328716/Test1.zip)

lochel:
Thanks for the report together with an example. I don't quite know why it fails, but it might be related to the connectors in the Root system. Renaming works fine if I first remove those. I will investigate more later today.

lochel:
The problem is connected to the internal data management within OMSimulator. It has nothing to do with OMEdit. The following lua script can be used for debugging:

```lua
oms_setCommandLineOption("--suppressPath=true")
oms_setTempDirectory("./Test1/")

model = oms_importFile("Test1.ssp")
print(model)

oms_rename(model, "Gru")

src = oms_exportSnapshot("Gru")
--print(src)
```

arun3688:
@max-privato The issue is fixed with this commit https://github.com/OpenModelica/OMSimulator/pull/1009, you can verify it

max-privato:
I tested with 1.18.0-dev-239 win 64 bit (24 April) and I still had a crash

arun3688:
@max-privato The fix is not merged with OpenModelica master yet, so you have to wait a little bit more

lochel:
Sorry for the delay. We will merge it within a few days and let you know.

lochel:
@max-privato this issue has been fixed as far as I can tell. Can you confirm, please?

max-privato:
Yes it is fixed!
Thanks a lot for fixing this (and the many tickets I'm opening on SSP...)

PR #1009, opened 2021-04-19
fix connectors when renaming model

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/1008

### Purpose

This PR fixes the `connectors` when renaming the `model`

### DIFF
--- a/src/OMSimulatorLib/Model.cpp
+++ b/src/OMSimulatorLib/Model.cpp
@@ -121,9 +121,11 @@ oms_status_enu_t oms::Model::rename(const oms::ComRef& cref)
   this->cref = cref;
 
   if (system)
+  {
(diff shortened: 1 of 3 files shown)


src/OMSimulatorLib/Model.cpp
```diff
@@ -119,6 +119,12 @@ oms_status_enu_t oms::Model::rename(const oms::ComRef& cref)
     return logError(std::string(cref) + " is not a valid ident");
 
   this->cref = cref;
+
+  if (system)
+    system->renameConnectors();
+    for (const auto & it:system->getSubSystems())
+      it.second->renameConnectors();
```
lochel, line 126: C++ is not Python 😄

Issue #1007, opened 2021-04-16
SSP use from OMEdit causes Infinite loop while showing messages


max-privato:
### Description

I have a Work-in-progress ssp. 
When I run it OMEdit enters an infinite loop in showing error messages and to see them I must get a screen print and copy somewhere. 
Even if the model has mistakes, this behaviour is unexpected.

Sorry for not creating a MWE, but I don't have time now. In case of need, I could try next week.

### Steps to reproduce the behavior

Load ErrorLoop.ssp and run it

### Version and OS

OM: 1.18-deb 224, Win10 64 bit

- Version: [`OMSimulator --version`]
- OS: [e.g. Win10, 64bit]

[ErrorLoop.zip](https://github.com/OpenModelica/OMSimulator/files/6322980/ErrorLoop.zip)

lochel:
@arun3688 can you please test it on Windows?

arun3688:
@lochel I tested and it is running into infinite loop with the following error

```
>>> OMSimulator ErrorLoop.ssp                                                                                                                                 
warning: GRU_TER (): fmi2SetupExperiment: tolerance control not supported for fmuType fmi2ModelExchange, setting toleranceDefined to fmi2False              
info:    maximum step size for 'GruTerOne.Root': 0.100000                                                                                                   
info:    Result file: GruTerOne_res.mat (bufferSize=10)                                                                                                     
                                                                                                                                                            
[CVODE ERROR]  cvInitialSetup                                                                                                                               
  Initial ewt has component(s) equal to zero (illegal).                                                                                                     
                                                                                                                                                            
error:   [doStep] SUNDIALS_ERROR: CVode() failed with flag = -22                                                                                            
                                                                                                                                                            
[CVODE ERROR]  cvInitialSetup                                                                                                                               
  Initial ewt has component(s) equal to zero (illegal).                                                                                                     
                                                                                                                                                            
error:   [doStep] SUNDIALS_ERROR: CVode() failed with flag = -22                                                                                            
                                                                                                                                                            
[CVODE ERROR]  cvInitialSetup                                                                                                                               
  Initial ewt has component(s) equal to zero (illegal).                                                                                                     
                                                                                                                                                            
error:   [doStep] SUNDIALS_ERROR: CVode() failed with flag = -22                                                                                            
                                                                                                                                                            
[CVODE ERROR]  cvInitialSetup                                                                                                                               
[...]

PR #1010, opened 2021-04-22
Check if nominals of FMU are greater then 0

AnHeuermann:
Check if nominals of FMU are greater then 0.
If they are not report an error and abort initialization.

### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/1007

### Purpose

When a FMU has illegal nominals (<=0) we report an error. Otherwise the absolute tolerances used for CVODE are 0 which will result in an endless loop.

### Approach

Just report an error and stop.


lochel (approved):
LGTM, thanks @AnHeuermann

Issue #1011, opened 2021-04-22
Connecting integer outputs (in modelica connector class) to enum input 


robha67:
### Description

OMsimulator reports the folowing error when a integer output on the form "<connectorName>.<signalName>" is connected to an enum:

"Type mismatch in connection:..."

### Steps to reproduce the behavior
Checkout "Embrace_trunk\WP3\D3.0\Saab\IntToEnumExample" from the embrace repository.
Run the file "SSP_Export.py" available in the folder mentioned above. 

### Version and OS

Linux, CentOS 7.4 VS. OMSimulator v2.1.1.post59-g0165678-linux

@lochel @arun3688

robha67:
It seems like the connection works if i change the order of the signals i.e
addConnection("model.root.AdaptionUnit.sWsignals.Aircraft_State","model.root.ECS_SW.Input.Aircraft_State"): this works
addConnection("model.root.ECS_SW.Input.Aircraft_State","model.root.AdaptionUnit.sWsignals.Aircraft_State"): this produces the error above

The order does not matter for reals that are connected in the same way

lochel:
@robha67 thanks for the example. I was now able to make a minimal example for debugging, see OpenModelica/OpenModelica#7404.

lochel:
This is fixed with #1012.

PR #1012, opened 2021-04-23
fix integer connections to enumerations

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/1011

### Purpose

This PR fixes integer connections to enumerations types

### DIFF
--- a/src/OMSimulatorLib/System.cpp
+++ b/src/OMSimulatorLib/System.cpp
@@ -1087,7 +1087,7 @@ oms_status_enu_t oms::System::addConnection(const oms::ComRef& crefA, const oms:
   {
     // allow non-real connections to tables
   }
-  else if(conA->getType() == oms_signal_type_integer && conB->getType() == oms_signal_type_enum)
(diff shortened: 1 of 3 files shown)


src/OMSimulatorLib/System.cpp
```diff
@@ -1087,7 +1087,7 @@ oms_status_enu_t oms::System::addConnection(const oms::ComRef& crefA, const oms:
   {
     // allow non-real connections to tables
   }
-  else if(conA->getType() == oms_signal_type_integer && conB->getType() == oms_signal_type_enum)
+  else if((conA->getType() == oms_signal_type_integer || conA->getType() == oms_signal_type_enum)  && (conB->getType() == oms_signal_type_enum || conB->getType() == oms_signal_type_integer))
```
lochel, line 1090: Shoudln't it be like this?
```else if((conA->getType() == oms_signal_type_integer && conB->getType() == oms_signal_type_enum) || (conA->getType() == oms_signal_type_enum && conB->getType() == oms_signal_type_integer))```

arun3688: this is fixed with  https://github.com/OpenModelica/OMSimulator/pull/1012/commits/bb23e8db430311948b76c3df81655b5436808a8d

lochel (approved):
Thanks!

PR #1016, opened 2021-04-26
Updated embrace use case

robha67:
### Related Issues
Updated use-case with a simulink generated SW FMU. The use-case now include "bus-to-bus" type connections including a int to enum connection. 

### Purpose
Open test platform capturing end-user requirements on OMSimulator


lochel:
Sorry, this was closed by accident.

lochel:
@robha67 Sorry that it took so long, but now the example is adapted to the test suite setup and should hopefully pass all checks.

lochel:
@robha67 I had to reduce the stopTime from 1200 to 12 in order to make the test pass in reasonable time. Please let me know if you see any problem with that.

robha67:
I could imagine that the longer time step could cause some instability problems and erranous results. Is the OM simulator checked for numerical stability? If not, then I don't see a problem. @lochel

lochel:
I didn't change the step size. The simulation just terminates at stopTime=12 instead of stopTime=1200 because the test case otherwise takes way too much time.

robha67:
Oh, ok. Then I missunderstood you. I see no reason to run the entire mission. I guess a longer simulation is needed to se the effect concerning the potential issue of increasing overhead but I believe that investigation is best done "off-line"

PR #1032, opened 2021-06-11
The atomic library needs to be linked on more Linux versions

sjoelund:
Previously only 32-bit Linux were linked, but arm64 (on Xenial, but not
more recent versions) also needs it.


adrpo (approved):
LGTM!

lochel (approved):

lochel:
@sjoelund await the requested reviews before merging!

PR #1035, opened 2021-06-12
export unique ssp

arun3688:
### Purpose

This PR fix tests to export unique ssp's


lochel (approved):
The test is not testing much though, right?

arun3688:
> The test is not testing much though, right?

Yes it does not test much, just to avoid conflicts with jenkins testing

lochel:
I meant the original test which is just checking the return value and not the actual export.

lochel:
I meant the original test which is just checking the return value and not the actual export.

Issue #789, opened 2020-08-14
OMSimulator can't handle long relative path on Windows


AnHeuermann:
### Description
When trying to simulate a FMU from a different directory I get various errors from OMSimulator.
When simulating it from the same directory (or only a few directories away) it works.

### Steps to reproduce the behavior
I'm working inside the [FMI Cross-Check repository](https://github.com/modelica/fmi-cross-check) in a 64 bit mingw shell on Windows 10 (get OMDev from git: `$git clone https://openmodelica.org/git/OMDev.git`).

Make sure `ls` can find the FMU:
```bash
/d/workspace/OpenModelica/OMSimulator/testsuite/fmi-cross-check/fmi-cross-check/results/2.0/me/win64/OMSimulator/v2.1.0/SimulationX/4.0.4/ControlledTemperature
$ ls "../../../../../../../../../fmus/2.0/me/win64/SimulationX/4.0.4/ControlledTemperature/ControlledTemperature.fmu"
../../../../../../../../../fmus/2.0/me/win64/SimulationX/4.0.4/ControlledTemperature/ControlledTemperature.fmu
```

#### Simulate the FMU directly:
```bash
$ OMSimulator "../../../../../../../../../fmus/2.0/me/win64/SimulationX/4.0.4/ControlledTemperature/ControlledTemperature.fmu"
info:    Set temp directory to    "D:/workspace\OpenModelica\OMSimulator\testsuite\fmi-cross-check\fmi-cross-check\results\2.0\me\win64\OMSimulator\v2.1.0\SimulationX\4.0.4\ControlledTemperature"
info:    Set working directory to "D:/workspace\OpenModelica\OMSimulator\testsuite\fmi-cross-check\fmi-cross-check\results\2.0\me\win64\OMSimulator\v2.1.0\SimulationX\4.0.4\ControlledTemperature"
info:    Set working directory to "D:/workspace\OpenModelica\OMSimulator\testsuite\fmi-cross-check\fmi-cross-check\results\2.0\me\win64\OMSimulator\v2.1.0\SimulationX\4.0.4\ControlledTemperature"
error:   [oms_extractFMIKind] failed to extract modelDescription.xml from "../../../../../../../../../fmus/2.0/me/win64/SimulationX/4.0.4/ControlledTemperature/ControlledTemperature.fmu"
error:   [oms_RunFile] oms_extractFMIKind failed
info:    0 warnings
info:    2 errors
```

#### Simulate with a lua script
```lua
[...]

PR #1038, opened 2021-07-05
Use absolute path to check if FMU exists.

AnHeuermann:
Solves errors like
```bash
error:   [addSubModel] file does not exist: "..\..\..\..\..\..\..\..\..\fmus\2.0\cs\win64\Dymola\2019FD01\ControlledTemperature\ControlledTemperature.fmu"
```
on Windows 10 or

```bash
error:   [fmiLogger] module FMICAPI: Could not load the FMU binary: File name is too long.
error:   [instantiate] Could not load "resources/0001_fmu.fmu" which is associated with "model.root.fmu"; it may be corrupted or may not support your platform
```
on Ubuntu with wine. Example is from https://libraries.openmodelica.org/fmi-crosschecking/OMSimulator/history/build-v2.1.1-58%202021-03-25/wine64/files/20.cs.Dymola.2019FD01.ControlledTemperature/20.cs.Dymola.2019FD01.ControlledTemperature.log.html

### Related Issues

Partial fix for https://github.com/OpenModelica/OMSimulator/issues/789, but if the absolute path is still to long there is not much I can do about it. The function in question is `std::filesystem::exists`.

### Purpose

Fixes 

### Approach

Convert all paths passed to `addSubModel` to an absolute path before doing anything with it.


lochel (approved):

AnHeuermann:
IT could be possible for users to change their Windows to accept longer paths: https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd

But other then that I can't see much more we can do to solve to long names.

AnHeuermann:
It seems that the test is stuck: https://test.openmodelica.org/jenkins/blue/organizations/jenkins/OMSimulator/detail/PR-1016/7/pipeline

lochel:
@AnHeuermann please have a look. It's not very sophisticated, but might be just good enough.

PR #1049, opened 2021-07-07
Fix step sizes for cvode in sc-systems

lochel:


testsuite/fmi-cross-check/importFMUs.py
```diff
@@ -28,11 +28,12 @@ def generateLua(modelName, testFMUDir, resultDir, fmiType):
   luaFilePath = os.path.join(resultDir, modelName + ".lua")
 
   # Set OMSimulator settings
-  tempDir = "temp"
+  tempDir = "/temp/cross-check/"
```
lochel, line 31: @AnHeuermann This might solve the issue with too long file names, right? Maybe we have to delete the temp folders in this path from time to time... any suggestions?

testsuite/fmi-cross-check/importFMUs.py
```diff
@@ -160,9 +160,6 @@ def simulateFMU(omsimulator, testFMUDir, resultDir, modelName, fmiType, luaFile)
   if (not os.path.isfile(os.path.join(resultDir,  modelName+"_out.csv"))) and (exitCode == 0):
     exitCode = 1
 
-  # Delete temp files
-  shutil.rmtree(tempDir, ignore_errors=True)
-
```
lochel, line 165: @AnHeuermann Is this required?

testsuite/fmi-cross-check/importFMUs.py
AnHeuermann: This probably won't work on Windows, but should be fine for some Unix systems. We could just use some Python function to get the temp files directory, e.g the way described in https://stackoverflow.com/questions/847850/cross-platform-way-of-getting-temp-directory-in-python.

testsuite/fmi-cross-check/importFMUs.py
AnHeuermann: No, it's just added to make it more easy to create the PR for the Cross-Check repo. BUt nothing what a `find` from shell can't solve.

AnHeuermann (changes requested):
Using `tempfile.gettempdir()` should make the approach working on most OS.

testsuite/fmi-cross-check/importFMUs.py
AnHeuermann: Ah and to make a rerun possible. But I was using `git clean` anyway, just to be super sure that the working directory is clean.

testsuite/fmi-cross-check/importFMUs.py
lochel: Good point 👍

testsuite/fmi-cross-check/importFMUs.py
lochel: git clean will not touch the temp directory though

lochel: But OMSimulator is (or should be) cleaning its own garbage. So it should be fine I guess.

PR #1051, opened 2021-07-09
Reset default maximumStepSize for cross-check

lochel:


testsuite/fmi-cross-check/importFMUs.py
```diff
@@ -54,6 +53,7 @@ def generateLua(modelName, testFMUDir, resultDir, fmiType):
     if not df.loc["AbsTol", 1] == 0:
       absTol = str(df.loc["AbsTol", 1])
 
+  maximumStepSize = str((float(stopTime) - float(startTime)) / 500)
   if "StepSize" in df.index:
     if not df.loc["StepSize", 1] == 0:
       maximumStepSize = str(df.loc["StepSize", 1])
```
AnHeuermann, line 59: Should we revert
```
f.write("oms_setVariableStepSize(\"model\", 1e-12, 1e-12, " + maximumStepSize + ")\n")
```
as well? But a smaller possible minimum step size sound reasonable.

AnHeuermann (commented):
Looks good. Not sure about the minimus step size in `oms_setVariableStepSize`. But I think it doesn't change anything.

testsuite/fmi-cross-check/importFMUs.py
lochel: 1e-12 was used before - it was just hard-coded in OMSimulator and the value provided via API was ignored. So I guess we can keep it for now.

PR #1061, opened 2021-07-12
Command line option for the step size

lochel:


src/OMSimulatorLib/Flags.h
```diff
@@ -158,6 +165,7 @@ namespace oms
       {"--solver", "", "Specifies the integration method (euler, [cvode])", re_default, Flags::Solver, false},
       {"--solverStats", "", "Adds solver stats to the result file, e.g. step size; not supported for all solvers (true, [false])", re_bool, Flags::SolverStats, false},
       {"--startTime", "-s", "Specifies the start time", re_double, Flags::StartTime, false},
+      {"--stepSize", "", "Set step size", re_stepSize, Flags::StepSize, false},
```
AnHeuermann, line 168: Add a hint on how to use `--stepSize` with minimum and maximum step size.

src/OMSimulatorLib/Flags.cpp
```diff
@@ -391,6 +396,32 @@ oms_status_enu_t oms::Flags::StartTime(const std::string& value)
...
+
+  return result;
+}
+
+oms_status_enu_t oms::Flags::StepSize(const std::string& value)
+{
+  std::vector<std::string> stepSize = split(value, ',');
+  if (stepSize.size() > 1)
+  {
+    GetInstance().initialStepSize = atof(stepSize[0].c_str());
```
AnHeuermann, line 416: We could add some basic error checks, e.g. no negative numbers for steps sizes.
But that was the only way I could find to break this flag.

src/OMSimulatorLib/Flags.h
lochel: done.

src/OMSimulatorLib/Flags.cpp
lochel: done

Issue #951, opened 2021-02-16
Functionality to update exsting SSV file with paramter values


robha67:
Functionality enabling the modification of, and the generation of, separate SSV files. The functionality should allow the user to export parameter values specified via, for example, OMS.setReal without generating a complete SSP. 
Then, it should be possible to, via some API command, to invoke such an SSV file on an existing SSP i.e. add it to the resource folder and point to it in the SSD. When adding the "new" SSV file to the resources of the SSP, previously existing SSV files should not be overwritten. It should be possible to change between several SSV files (corresponding to the same model) located in the resources folder. @arun3688

arun3688:
@robha67 I think the first feature, generating of SSV files with update start values, should look something like below

```
oms_exportSSVTemplate("model") // list of all default start values in modeldescription.xml
oms_setReal("model.root.A", 10) // set new start values
oms_setReal("model.root.B", 20) // set new start values
oms_exportSSVTemplate("model")  // should update the new start values in the generated ssv template 
```
The second feature , having a new API to insert the new ssv file in the `ssp` package is basically unzipping and zipping and i guess this can be done by yourself or we can do it via an API also @lochel  can you comment on this?

And the third feature, switching between different SSV files, I think this can be done via the `oms_importFile()` API with some modification like

```
oms_importFile("import_export_parameters_inline.ssp"); // default ssv files will be used 
oms_importFile("import_export_parameters_inline.ssp", "config1.ssv")  // while importing we can override the default file with config1.ssv if it exisits in the resource folder otherwise default `.ssv file` in ssd file will be used.
oms_importFile("import_export_parameters_inline.ssp", "config2.ssv") 
```

I hope you want something like this ? 

@lochel can you give some suggestions or comments ?

robha67:
1)

**_I think the first feature, generating of SSV files with update start values, should look something like below

oms_exportSSVTemplate("model") // list of all default start values in modeldescription.xml
oms_setReal("model.root.A", 10) // set new start values
oms_setReal("model.root.B", 20) // set new start values
oms_exportSSVTemplate("model")  // should update the new start values in the generated ssv template_** 

Perfect solution, exactly what we need. 

2)
The second feature , having a new API to insert the new ssv file in the ssp package is basically unzipping and zipping and i guess this can be done by yourself or we can do it via an API

[...]

PR #1079, opened 2021-08-19
delete parameter bindings references in ssd

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/951

### Purpose

This PR provides functionality to switch between different ssv files in SSP. The following operations should be supported

- [x] deleting only references in ssp 
- [x] deleting both references and resources files in ssp
- [x] adding new resources to an existing ssp and reference it in ssd

### DIFF
--- a/src/OMSimulatorLib/Component.h
+++ b/src/OMSimulatorLib/Component.h
@@ -91,7 +91,7 @@ namespace oms
     virtual oms_status_enu_t setReal(const ComRef& cref, double value) { return logError_NotImplemented; }
     virtual oms_status_enu_t setRealInputDerivative(const ComRef& cref, const SignalDerivative& der) { return logError_NotImplemented; }
     virtual oms_status_enu_t stepUntil(double stopTime) { return oms_status_ok; }
-    virtual oms_status_enu_t newResources(const std::string& filename) { return logError_NotImplemented; }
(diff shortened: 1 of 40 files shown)


src/OMSimulatorLib/Component.h
```diff
@@ -92,6 +92,8 @@ namespace oms
     virtual oms_status_enu_t setRealInputDerivative(const ComRef& cref, const SignalDerivative& der) { return logError_NotImplemented; }
     virtual oms_status_enu_t stepUntil(double stopTime) { return oms_status_ok; }
     virtual oms_status_enu_t addResources(std::string& filename) { return oms_status_ok; }
+    virtual oms_status_enu_t deleteReferencesInSSD(const std::string& filename) {return oms_status_ok;}
+    virtual oms_status_enu_t deleteResourcesInSSP(const std::string& filename) {return oms_status_ok;}
```
lochel, line 96: Should we return oms_status_error and a message that these functions aren't implemented?

arun3688: but these functions are implemented in `ComponentFMUCS.cpp` and `ComponentFMUME.cpp`

arun3688: this is fixed with this commit https://github.com/OpenModelica/OMSimulator/pull/1079/commits/4b004e3227d4e44b7804db1f8a8010131cde74d2

PR #1084, opened 2021-09-06
Embrace dt 2.1

robha67:
### Purpose

Updated Saab Embrace use-case with a bug fix of the units in the SSV files. The SSP now also contains two different configurations. The updated SSP is placed, unpacked, in the resources folder. A new simulation script (embrace_test2Conf.py) is added to the AircraftVehicleDemonstrator folder. Please note that this script is not yet complete and tested. The switch between configuration files need to be added to the script.


robha67:
The two conflicts should not be there, the conflicts are unintentional and could be reverted.
@lochel @arun3688

PR #1086, opened 2021-09-07
Embracetwoconf

arun3688:
### Related Issues
https://github.com/OpenModelica/OMSimulator/pull/1084

### Purpose
Updated Saab Embrace use-case with a bug fix of the units in the SSV files. The SSP now also contains two different configurations. The updated SSP is placed, unpacked, in the resources folder. A new simulation script (embrace_test2Conf.py) is added to the AircraftVehicleDemonstrator folder.


lochel:
@arun3688 What is this about? Is it obsolete or should we try to integrate the changes?

Issue #900, opened 2020-12-18
Update OMSimulator graphical modeling documentation 


AnHeuermann:
### Description

The [documentation for graphical modeling with OMSimulator](https://openmodelica.org/doc/OMSimulator/master/html/OMEdit.html#new-omsimulator-model) from OMEdit is using old screenshots.
When using a recent nightly build of OMEdit you don't have a menu item `OMSimulator` to create a new OMSimulator model.

### Expected behavior

I only see an option to create new SSP models. So it should be described how to create some models with that.

### Screenshots

<img width="443" alt="Screenshot_OMEdit" src="https://user-images.githubusercontent.com/38031952/102611616-8aa0e280-412f-11eb-87c1-ea7371c64851.png">

### Version and OS

- Version: OMEdit connected to OpenModelica v1.17.0-dev-281-g1a53775ae8 (64-bit)
and to OMSimulator v2.0.0.post259-g2d9666d-mingw
- OS: Win10, 64bit

AnHeuermann:
I'm going to update it today.

lochel:
Thanks @AnHeuermann!

AnHeuermann:
The changes are now life in the [OMSimulator User's Guide](https://openmodelica.org/doc/OMSimulator/master/html/OMEdit.html#), but not the OpenModelica User's Guide yet. But they will synchronize when the OMSimulator sub-repo is updated in OpenModelica/OpenModelica.

PR #1087, opened 2021-09-08
Updating UsersGuide on graphical modeling with SSP

AnHeuermann:
### Related Issues

#900 

### Purpose

Describe how to do SSP modeling with OMEdit and OMSimulator.


AnHeuermann:
Check out `html/OMEdit.htlm` from the User's Guide artifact in the Jenkins job.

lochel (approved):
LGTM, thanks!

lochel:
Would be nice to show some results on the last figure next to the composite model. But I think even without it is a good improvement and should be merged.

PR #1097, opened 2021-11-17
Include cmath explicitly.

mahge:
### Purpose

`cmath` is not included directly.

  - The files use `std::isnan` and `std::isinf` from `cmath`.
  - It breaks on old compilers (e.g. gcc-5) because the header might not
    be included indirectly.

### Approach
Include `cmath` explicitly.

### DIFF
--- a/src/OMSimulatorLib/ComponentFMUCS.cpp
+++ b/src/OMSimulatorLib/ComponentFMUCS.cpp
@@ -44,6 +44,8 @@
 #include <JM/jm_portability.h>
 #include <RegEx.h>
 #include <unordered_set>
+#include <cmath>
+
 
 oms::ComponentFMUCS::ComponentFMUCS(const ComRef& cref, System* parentSystem, const std::string& fmuPath)
   : oms::Component(cref, oms_component_fmu, parentSystem, fmuPath), fmuInfo(fmuPath)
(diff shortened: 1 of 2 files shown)


lochel (approved):
Thanks @mahge

lochel:
I just restarted the Jenkins job and hopefully it will pass just fine this time.

mahge:
Hmm. I think I fetched master and rebased it properly. So it is probably not anything missing from master.

There is something else going on. I will check again.

arun3688:
@mahge  The failing tests are related to TLM simulation, these are random failures occuring at jenkins, You can trigger the build again by using empty commit `git commit --allow-empty -m "Trigger notification"`

lochel:
I re-started the Jenkins job a few times. If it is just random failures then things got much more unstable recently.

sjoelund:
The biggest change made to Jenkins is more machines run Ubuntu 20.04 instead of 18.04 now. It's still docker images, but if something depends on the kernel that could influence things.

lochel:
I don't what to do except for re-running the job again and again. @sjoelund do you have a better idea?

sjoelund:
Considering the library coverage using OMSimulator also hangs and times out, I would guess there is a race condition somewhere. Loading TLM dynamically only on demand might help. Not using network connections might also help since ports are a blocking resource.

PR #1100, opened 2021-11-26
Include cmath explicitly.

mahge:
This is a "retry" of #1097. To see if there is some transient issue with the PR itself.

### Purpose

`cmath` is not included directly.

  - The files use `std::isnan` and `std::isinf` from `cmath`.
  - It breaks on old compilers (e.g. gcc-5) because the header might not
    be included indirectly.

### Approach
Include `cmath` explicitly.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -728,6 +728,7 @@ void partest(cache=true, extraArgs='') {
 
   sh ("""#!/bin/bash -x
   ulimit -t 1500
+  ulimit -v 6291456 # Max 6GB per process
 
   cd testsuite/partest
   ./runtests.pl ${env.ASAN ? "-asan": ""} ${env.ASAN ? "-j1": "-j${numPhysicalCPU()}"} -nocolour ${env.BRANCH_NAME == "master" ? "-notlm" : ""} -with-xml ${params.RUNTESTS_FLAG} ${extraArgs}
--- a/src/OMSimulatorLib/ComponentFMUCS.cpp
+++ b/src/OMSimulatorLib/ComponentFMUCS.cpp
@@ -44,6 +44,8 @@
 #include <JM/jm_portability.h>
 #include <RegEx.h>
 #include <unordered_set>
+#include <cmath>
+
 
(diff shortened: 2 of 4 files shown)


lochel:
It keeps failing unfortunately. @adrpo do you have an idea how to get things back running again?

adrpo:
Looking at `ps aux | grep hudson` on ripper2 seems this one is stuck:

```
/usr/bin/perl -w ../rtest -v -nolib tlm3d.lua
/var/lib/jenkins1/ws/OMSimulator_PR-1100//install/linux/bin//OMSimulator tlm3d.lua
```

I'll gdb it to see where is stuck.

lochel:
Okay, otherwise we might have to disable that particular tlm test?

adrpo:
That test is using 20 terrabytes of memory?!?!!?! Geez :)

`1361045 hudson    20   0   20.0t   1.0g   8480 S 159.5   1.6  20:55.28 /var/lib/jenkins/ws/OMSimulator_PR-1100//install/linux/bin//OMSimulator tlm3d.lua`

I can't gdb to it, it won't let me. Maybe one could run valgrind on it to see where it goes totally wrong.

adrpo:
I killed the test as it used all the memory from that host.

sjoelund:
I wonder if it's related to the tests hanging on the library coverage runs.

Also, is there no ulimit for virtual memory?

adrpo:
I think is the asan version of the test as far as I can see.

adrpo:
No idea about the limits, but it seems no.

adrpo:
After I added `ulimit -v` to 6GB a lot of tests are failing!?

adrpo:
It really should not need that much memory I guess.

adrpo:
Any idea why this happens? Is asan using that much memory?

mahge:
The issue has been "worked-around" for now by disabling address sanitizer tests by default with #1106. So the original PR #1097 is now merged.

This can be closed or kept open until the issue is resolved for good.

Issue #1101, opened 2021-12-01
Building OMSimulator from OpenModelica master project broken on Windows.


AnHeuermann:
### Description

Building OMSimulator from the super Project OpenModelica/OpenModelica doesn't work on Windows (OMDev + mingw64).

### Steps to reproduce the behavior

```bash
cd /path/to/OpenModelica/
# make -f Makefile.omdev.mingw omsimulator -j14 --output-sync &> omsimulator.log
```

```
clang++  -std=c++11  -DUSE_THREADS  -O2 -I. -I../3rdParty/misc/include -I../3rdParty/rtime  -I/c/OMDev/tools/msys/mingw64/include/libxml2 -I/c/OMDev/tools/msys/mingw64/include/ -c Communication/TLMManagerComm.cc -o WINDOWS64/Communication/TLMManagerComm.o
Communication/TLMManagerComm.cc:26:9: warning: 'NOMINMAX' macro redefined [-Wmacro-redefined]
#define NOMINMAX
        ^
C:\OMDev\tools\msys\mingw64\include\c++\10.2.0\x86_64-w64-mingw32\bits/os_defines.h:45:9: note: previous definition is here
#define NOMINMAX 1
        ^
Communication/TLMManagerComm.cc:75:5: error: no matching function for call to 'setsockopt'
    setsockopt(theSckt, SOL_SOCKET, SO_REUSEADDR, &val, sizeof(int));
    ^~~~~~~~~~
C:\OMDev\tools\msys\mingw64\x86_64-w64-mingw32\include\winsock2.h:1029:34: note: candidate function not viable: no known conversion from 'int *' to 'const char *' for 4th argument
  WINSOCK_API_LINKAGE int WSAAPI setsockopt(SOCKET s,int level,int optname,const char *optval,int optlen);
                                 ^
1 warning and 1 error generated.
```

### Expected behavior

Compile all the things.

### Version and OS

- Version:
  - OpenModelica at 14546fe3d955fab7d2b2ff93f9a5b746c8fc31fe
  - OpenModelica/OMSimulator at 4c7059bbd55b6bd0fa2870375390a44222056074
  - OpenModelica/OMSimulator/OMTLMSimulator at e803cec2110ef470515cda3521e35a2f7b1173cc
  - OMDev at f5761810321b22a7d5280ee38c6d83f97052ec3f
- OS: Windows 10 with OMDev and MINGW64

AnHeuermann:
Changing

```C
[...]

PR #1102, opened 2021-12-01
Issue 1101

AnHeuermann:
### Related Issues

#1101

### Purpose

Resolve compiler errors with `setsockopt` on Windows.
Also fix some compiler warnings.

### DIFF
--- a/src/OMSimulatorLib/TLM/TLMBusConnector.h
+++ b/src/OMSimulatorLib/TLM/TLMBusConnector.h
@@ -19,24 +19,24 @@ namespace oms
 class Component;
 class System;
 
-typedef struct  {
+typedef struct oms_tlm_sigrefs_signal_t {
     int y = 0;
 } oms_tlm_sigrefs_signal_t;
 
-typedef struct  {
+typedef struct oms_tlm_sigrefs_1d_t {
     int x = 0;
     int v = 1;
     int f = 2;
 } oms_tlm_sigrefs_1d_t;
 
-typedef struct  {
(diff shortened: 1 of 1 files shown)


AnHeuermann:
Did I introduce a new memory leak in https://github.com/OpenModelica/OMTLMSimulator?

AnHeuermann:
@adrpo Do you know if the CI fails because of https://github.com/OpenModelica/OMTLMSimulator/pull/130 or because of https://github.com/OpenModelica/OMTLMSimulator/pull/131?

AnHeuermann:
The linux64-asan tlm3d.lua test was stuck (around an hour running and not stopped) in the last two Jenkins runs. I tried it with the same docker image on my machine a few times and it was always working. It is by far the slowest test (~50 seconds).

adrpo:
The problem is that asan consumes all memory (20t !!) on the machine and everything breaks, see discussion in #1100. When I limit the memory to 6GB *all* tests are failing.

adrpo:
Is strongly suggest to ulimit -v to 10gb or so max and maybe disable the asan tests.

sjoelund:
It's only virtual memory, so it's not actually using 20T. Only pages that are written to start using actual memory. But that's the way ASAN works. If you limit memory it will fail, no matter what limit under 20T... Maybe use valgrind instead?

AnHeuermann:
Okay, so now I understand why the runs happen to break so often.

I don't think there is any point in checking for memory problems in our CI if it breaks it every second run.

PR #1111, opened 2022-01-17
test windows fmus causing segmentation fault

arun3688:
### Purpose

This PR tests windows fmu generated from OpenModelica, This is a test PR which should check windows fmus are working fine in OMSimulator


arun3688:
the fmus are working correctly after doing make clean and then build the omc again

PR #1116, opened 2022-02-22
Add example equationPair

lochel:
There are only linux binaries for the example right now. @AnHeuermann could generate a cross-platform FMU for us? All the docker instructions are in `equationPair.mos`.


testsuite/resources/equationPair.mos
```diff
@@ -0,0 +1,9 @@
+setCommandLineOptions("-d=-disableDirectionalDerivatives,fmuExperimental"); getErrorString();
+
+loadFile("equationPair.mo"); getErrorString();
+
+buildModelFMU(equationPair.equation1, version="2.0", fmuType="cs", fileNamePrefix="equationPair.equation1", platforms={"i686-linux-gnu docker run docker.openmodelica.org/build-deps:v1.13-i386","x86_64-linux-gnu docker run docker.openmodelica.org/build-deps:bionic.nightly.amd64","i686-w64-mingw32 docker run docker.openmodelica.org/msyscross-omsimulator:v2.0","x86_64-w64-mingw32 docker run docker.openmodelica.org/msyscross-omsimulator:v2.0"}); getErrorString();
```
AnHeuermann, line 5: OMEdit/ omc uses the `multiarch/crossbuild` docker image. I guess we should use that here as well:

```
platforms := {
  "x86_64-apple-darwin15 docker run -e CROSS_TRIPLE=x86_64-apple-darwin multiarch/crossbuild",
  "arm-linux-gnueabihf docker run -e CROSS_TRIPLE=arm-linux-gnueabihf multiarch/crossbuild",
  "x86_64-linux-gnu docker run multiarch/crossbuild",
  "i686-linux-gnu docker run multiarch/crossbuild",
  "x86_64-w64-mingw32 docker run -e CROSS_TRIPLE=x86_64-w64-mingw32 multiarch/crossbuild",
  "i686-w64-mingw32 docker run -e CROSS_TRIPLE=i686-w64-mingw32 multiarch/crossbuild"};
```

sjoelund: In that case you would need to use a specific tag that never updates since otherwise the tests might start to fail randomly since our machines might use different versions of the image.

lochel: The mos script isn't part of the actual test.

PR #1134, opened 2022-03-04
Add model.getAllSignals to python api

lochel:


src/OMSimulatorPython/Model.py
```diff
@@ -56,6 +59,17 @@ def reset(self):
     if Types.Status(status) != Types.Status.OK:
       raise Exception('error {}'.format(Types.Status(status)))
 
+  def getAllVariables(self):
```
lochel, line 62: Maybe change it to getAllSignals

Issue #1138, opened 2022-03-08
Cannot update variable start value in instantiated state


bilderbuchi:
### Description
It does not seem to be possible to set the value of a state variable in an instantiated fmu.

The [docstring ](https://www.openmodelica.org/doc/OpenModelicaUsersGuide/latest/omsimulator.html#id143)of `setReal` (Python API) says: 
> Sets the value of a given real signal.
>
> status = oms.setReal(cref, value)
>
> This function can be called in different model states:
>
 >   Before instantiation: setReal can be used to set start values or to define initial unknowns (e.g. parameters, states). The values are not immediately applied to the simulation unit, since it isn't actually instantiated.
>
>    After instantiation and before initialization: Same as before instantiation, but the values are applied immediately to the simulation unit.
>
>    After initialization: Can be used to force external inputs, which might cause discrete changes of continuous signals.

It seems that contrary to that, in the state after instantiation and before initialization, this is currently broken, as I can't seem to set values.  
Also contrary to the docstring, if applied before instantiation, the value seems to be _in fact immediately applied_.

### Steps to reproduce the behavior

```modelica
model Issue_FMU_update_vars
  Real var;
  Real var_with_start_nonfixed(start=2.2, fixed=false);
  Real var_with_start_fixed(start=4.4, fixed=true);
  Real var_with_start_param(start=startparameter, fixed=true);
  parameter Real startparameter = 6.6;
equation
  der(var) = 1.0;
  der(var_with_start_nonfixed) = 2.0;
  der(var_with_start_fixed) = 4.0;
  der(var_with_start_param) = 6.0;
end Issue_FMU_update_vars;
```
(Variables beside `var` were just used for exploring the issue)

```python
import OMSimulator as oms

fmu_location = 'Issue_FMU_update_vars.fmu'
model = oms.newModel("model")
root = model.addSystem('root', oms.Types.System.SC)
fmu = root.addSubModel('fmu', fmu_location)
print(f'Model state: {model.modelState}')
print(f'Initial value: {fmu.getReal("var")}')
fmu.setReal('var', 11.0)
print(f'After setReal: {fmu.getReal("var")} - works but wasteful to always destroy a model for repeated runs')

model.instantiate()
print(f'Model state: {model.modelState}')
print(f'After instantiate: {fmu.getReal("var")}')
fmu.setReal('var', 22.0)
print(f'After setReal: {fmu.getReal("var")} - value did not "take" in instantiated state')

model.initialize()
[...]

PR #1140, opened 2022-03-10
Fix variable start value in instantiated state

lochel:
### Related Issues

- #1138


arun3688:
@this is fixed with this commit https://github.com/OpenModelica/OpenModelica/commit/c9ba3f488006dcbfbdc612e9116ffd8222532b7d

Issue #942, opened 2021-02-11
Automatic unit conversion (SSP)


robha67:
Functionality for automated conversion of units between SSV parameters and FMU parameters.

If possible, the conversion should be stored in the SSD. Otherwise, the SSP is unusable outside of OMSimulator. I can't find that this is allowed in the specification. Maybe there is some workaround? Could a "conversion model" be added to the the SSP via some OMSimulator API command? Could maybe a signal dictionary be exploited for such a "conversion model"?

A different, and possibly simpler approach, would be to implement some API command that scans the SSV values, converts any non-SI units to SI, and saves the updated SSV.

lochel:
What sort of conversions do you have in mind? Millimetre to metre? Or something more complicated?

robha67:
Yes. Mm to m, g to kg, etc. The units expressed in an ssv could maybe be compared to the corresponding unit in the modelDescription via an ssm file

@lochel @arun3688

arun3688:
@robha67 We are starting to work on this issue,  it would be good if you can provide some examples or inputs to the unit conversion

robha67:
Hmm, I don't have an example ready for use but I can fix something simple.

robha67:
Ok, so now I rememeber the origin of this feature request. In the [embrace_TwoConf] test case, the lengths in the SSV are originally expressed in mm instead of m. So, replacing rows 1543 to 1545 with something like 

<ssv:Parameter name="part_feedLineRadar.parameterSet_outputParameters.parameter_pipeLengthTotalFL">
      <ssv:Real unit="[mm]" value="4614.34420887414"/>
</ssv:Parameter>

would reproduce the original test case. I guess that the unit mm also needs to be added to the <Units>.
The FMU parameter specified by this value is in meters.

@arun3688 do you want me to update the test-case and make a new pull request or is it more convenient for you to do the addition to the SSP?

arun3688:
> <ssv:Parameter name="part_feedLineRadar.parameterSet_outputParameters.parameter_pipeLengthTotalFL"> <ssv:Real unit="[mm]" value="4614.34420887414"/> </ssv:Parameter>
> 
> would reproduce the original test case. I guess that the unit mm also needs to be added to the . The FMU parameter specified by this value is in meters.

@robha67 I guess the unit `mm` should be converted to `m` when connecting to fmu parameters , this is what you mean or do you specifying some other thing

robha67:
[...]

PR #1152, opened 2022-04-19
export units to connectors

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/942

### Purpose

This PR exports units to connectors, which will be used to check the units when making `connections` between the `connectors` 

### TODO

- [x] export units to connectors
- [x] import units from connectors
- [x] set units to connectors


lochel (approved):
🎉

Issue #1153, opened 2022-04-19
Variable names in results file, gets an whitespace added


Fredrik1Haider:
### Steps to reproduce the behavior

Reading the results file (an .csv) with python pandas.read_csv, the variable names contain an additional whitespace and quotation marks in front on the name, comes out as:

`'" model.root.anVariable"'`

Thus, requiring "filtering" to make it consistent with naming in ssm etc.

### Expected behavior

Comes out as:
`'model.root.anVariable'`

### Version and OS

- Version: [`OMSimulator --version`]
- Python: [v3.6]
- OS: [Linux 64bit]

arun3688:
@Fredrik1Haider the issue is fixed with this commit https://github.com/OpenModelica/OMSimulator/commit/86bee651f7eebbc1db55a0f6521edcc34478ee21, please test it

Fredrik1Haider:
@arun3688 tested and works as requested, thank you!

PR #1154, opened 2022-04-19
Remove leading whitespace for variable names in csv export

lochel:
### Related Issues

* #1153


lochel:
I think the quoting was done in case variable names contain a komma themselves. But that is probably not an issue, since it would be an invalid cref. @arun3688  Or should we check for tyhat case in only quote variable names if neccessary?

arun3688:
@lochel  we should remove the quotes in general

arun3688 (approved):

PR #1163, opened 2022-05-20
fix solver oms_mav and oms_mav2 in dostep

arun3688:
### Purpose

This PR fixes `return status` in solver `oms_mav` and `oms_mav2`


src/OMSimulatorLib/SystemWC.cpp
```diff
@@ -766,8 +768,10 @@ oms_status_enu_t oms::SystemWC::doStep()
 
     return oms_status_ok;
   }
+  else
+    return logError("Invalid solver selected");
 
-  return logError("Invalid solver selected");
+  return oms_status_error;
```
lochel, line 774: Can we keep this line as it was before?

arun3688: this is fixed with https://github.com/OpenModelica/OMSimulator/pull/1163/commits/aa3bd100a20c837d57980029e11f6bc80c372c84

lochel (approved):

Issue #1169, opened 2022-06-17
Architecture modelling in OMSimulator


arun3688:
**Is your feature request related to a problem? Please describe.**

 It should be possible to import an partial fmu with only modeldescription.xml consisting of variables with inputs and outputs and parameters etc.. and make connections with other fmus. And later develop this partial fmu to a working fmu and then integrate the working fmu in the ssp without losing the connections and other datas It is still possible that the newly developed fmus has more additional variables and also change in the naming of inputs and output variables, in such cases the connections with old variable names will be lost, The users must update the connections by themselves.  

**Describe the solution you'd like**
Implement a new API `replaceSubModel(const char* oldFmu, const char* newFmu)`, When replacing the fmu checks should be made in all ssp concepts like in `ssd, ssv and ssm`, so that connections and parameter settings are not lost. It is possible that the namings of inputs and parameters match, but the start values might have been changed, in such cases new start values will be applied in `ssd, ssv and ssm`.  In case the `Types` of `inputs and outputs and parameters`  differed then the variables are updated according to the new changes and the connections will be removed with warning messages to user

arun3688:
@robha67  Hi robert we are working on architectural modelling in `OMSimulator` and me and lennart had some discussion and decided to implement this feature, You can add your suggestion if you have anything needs to be implemented in this feature

robha67:
Hi! Can the three of us take a brief meeting regarding this, possibly next week? Virtually or IRL does not matter for me

arun3688:
@robha67 Lennart is in holidays, but i am available and we can decide the timings in email and if lennart is free he can join, otherwise we can have the meeting by ourselves

robha67:
@danlo76, can you test this?

lenaRB:
fixed

PR #1170, opened 2022-06-21
implement API oms_replaceSubModel()

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/1169

### Purpose

This PR implements a new API `oms_replaceSubModel()` which replaces an existing fmu with the new one, but keeps the connections if the replaced fmu has the same properties with the old fmus which should support the architectural modelling in OMSimulator

### TODO

- [x] add API oms_replaceSubModel()
- [x] add test cases
- [x] check the replaced component in  ssv and ssm files
- [x] add python support
- [x] add documentation


lochel:
Please preserve higher level definitions from the ssp file.

PR #1199, opened 2023-01-27
Improve cmake configuration

lochel:
### Purpose

Fixing the cmake configuration to make it easier to build and include as library into other projects.


lochel:
I want to include all the 3rdParty libraries directly into the cmake file, instead of building them in a separate step beforehand. @mahge do you agree?

lochel:
I guess we have to merge kinsol and cvode in the 3rdParty repository.

lochel:
Not sure why these errors still show up even though `FMILIB_BUILD_TESTS` is set to `FALSE`:
```
CMake Error at 3rdParty/FMIL/Config.cmake/runtime_test.cmake:58 (file):

  file COPY cannot find

  "/var/lib/jenkins1/ws/OMSimulator_PR-1199/3rdParty/FMIL/Test/try_to_uncompress_this_file.zip":

  No such file or directory.

Call Stack (most recent call first):

  3rdParty/FMIL/CMakeLists.txt:298 (include)





CMake Error at 3rdParty/FMIL/Config.cmake/runtime_test.cmake:63 (file):

  file COPY cannot find

  "/var/lib/jenkins1/ws/OMSimulator_PR-1199/3rdParty/FMIL/Test/try_to_compress_this_file.xml":

  No such file or directory.

Call Stack (most recent call first):

  3rdParty/FMIL/CMakeLists.txt:298 (include)
```

mahge:
> I want to include all the 3rdParty libraries directly into the cmake file, instead of building them in a separate step beforehand. @mahge do you agree?

Absolutely. That's what we do with the OpenModelica 3rdParty libraires and so far it has worked well and is quite flexible.

lochel:
The tlm part isn't supporting cmake yet.

lochel:
This was mainly to fix fmil which is now removed completely 😄 Thanks @arun3688 and @mahge!

PR #1212, opened 2023-04-21
Updating documentation

AnHeuermann:
### Related Issues

Realted to https://github.com/OpenModelica/OpenModelica/pull/10542

### Purpose

  - Fixing some warning for sphinx build of user's guide of OpenModelica

### DIFF
--- a/doc/UsersGuide/Makefile
+++ b/doc/UsersGuide/Makefile
@@ -1,7 +1,7 @@
 # Makefile for Sphinx documentation
 #
 
-OMSIMULATOR_ROOT = $(CURDIR)/../../
+OMSIMULATOR_ROOT = $(CURDIR)/../..
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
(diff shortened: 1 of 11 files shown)


AnHeuermann:
@arun3688 The OMSimulator User's Guide is broken for many API functions, where e.g. `#CAPTION#` is used, see e.g. https://openmodelica.org/doc/OMSimulator/master/html/api/setVariableStepSize.html.

We could remove those in this PR as well and fix the documentation.

AnHeuermann:
Looks like it. Maybe we add all .rst files at some place? What if we change them to be something like .rst.inc?

arun3688:
@AnHeuermann, this is happening with your changes ?

AnHeuermann:
The online documentation has an HTML file for each API function, see https://openmodelica.org/doc/OMSimulator/master/html/api/.
But I'm not sure why those files are generated in the first place. In `doc/UsersGuide/source/conf.py` only snippets from each api function.rst file are used.
I can try to change this in this PR as well.

PR #1214, opened 2023-04-25
Return error if CVODE failed

### LINKED ISSUE #1213: CVode error doesn't stop simulation / infinite loop
### Description

I have a broken FMU which can't satisfy error tests, because the evaluation of the ODE system is wrong (playing around with some neural networks in there). CVODE will fail the error test, OMSimulator prints a warning and then does the exact same step again. This loop never stops.

```Lua
oms_setTempDirectory("./temp/")
oms_newModel("model")
oms_addSystem("model.root", oms_system_sc)
oms_addSubModel("modelname", "../modelname.fmu")

oms_setStopTime("model", 1)
oms_setFixedStepSize("model.root", 2e-4)

oms_instantiate("model")
oms_initialize("model")
oms_simulate("model")
oms_terminate("model")
oms_delete("model")
```

```
[CVODE ERROR]  CVode
  At t = 1.0002 and h = 0.0002, the error test failed repeatedly or with |h| = hmin.
```

### Expected behavior

CVODE already tries multiple times (e.g. reducing the internal step size?) and fails to do the integrator step. In that case `oms::SystemSC::doStep` should fail after printing the error message.

### Version and OS

- Version: OMSimulator v2.1.1.post199-g2cd3a45-linux-debug
- OS: Ubuntu 22.04.2 LTS

AnHeuermann:
### Related Issues

Fixes  #1213.

### Purpose

  - Fix endless loop of broken FMUs where error test can't be satisfied.

### Approach

  - If integrator CVode in `oms::SystemSC::doStep()` fails return an error.


lochel:
@AnHeuermann, @arun3688 Could you provide an update on the status of this pull request?

lochel:
We can merge this one, but it would be nice to have a test case for it.

PR #1217, opened 2023-07-15
Update the docker images used for testing PRs to 1.16

mahge:
- This provides more recent versions of tools like CMake. It is also what is currently being used by the OpenModelica repo to test its own PRs.

  - This still might not provide a recent enough CMake version. We will see how it goes.

### DIFF
--- a/.CI/cache-32/Dockerfile
+++ b/.CI/cache-32/Dockerfile
@@ -1,5 +1,5 @@
 # Cannot be parametrized in Jenkins...
-FROM docker.openmodelica.org/build-deps:v1.13-i386
+FROM docker.openmodelica.org/build-deps:v1.16-i386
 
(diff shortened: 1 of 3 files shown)


mahge:
@lochel @arun3688 Do we still need to have 32 bit Linux builds of OMSimulator? 

I am trying to update the docker images used to build the PRs and it seems there are no recent 32 bit images available to use. The last one we have is tagged at 1.13 (which is what is being used by OMSimulator now). 

If there is no one actively using 32 bit Linux binaries of OMSimulator, perhaps it is a good time to retire them now.

mahge:
@lochel @arun3688 I have merged this in for now because there is a lot of other changes that are waiting on this. If the 32 bit Linux build is indeed needed, we can add an updated docker image for it and re-enable it later.

PR #1235, opened 2023-08-09
Use CMake directly in MSVC builds too.

mahge:


mahge:
Jenkins refuses to pass this PR. The build job itself does not seem to have any relevant issues. The testing job, however, hangs (never finishes). Maybe a new PR can make a difference.

PR #1248, opened 2023-08-25
Prune OMSimulator-3rdParty

lochel:
See OpenModelica/OMSimulator-3rdParty#76


mahge (approved):

mahge:
You can probably remove xerces as well. IIRC it was never used because it required a recent CMake version and the OpenModelica docker images did not provide it. Maybe now it is a good time to switch to it?

lochel:
IIRC we use xerces for certain xml checks when we read the modelDescription.xml file before actually loading an fmu. Maybe @arun3688 can comment on it.

arun3688:
@mahge me and lennart spoke about removing `xerces` and then we decided to use `xerces` for xml validation of fmus which will be really useful, so we should keep it as of now

PR #1276, opened 2023-11-29
Recover changes to build PyPI package

lochel:
This recoveres some changes to create a PyPI package for OMSimulator, but I'm uncertain about the current state of the changes. I would appreciate insights from @mahge on the best approach before proceeding further.

To generate the PyPI package, follow these steps:

1. Build the project using the `OM_OMS_ENABLE_PIP` option in CMake:

   ```bash
   cmake -S . -B build/ -DCMAKE_INSTALL_PREFIX=install/ -DOM_OMS_ENABLE_PIP=ON
   cmake --build build/ --target install
   ```

2. Run the Makefile:

   ```bash
   make pip
   ```

3. Upload the package using twine:

   ```bash
   twine upload src/pip/install/dist/OMSimulator-2.1.1.post248.tar.gz
   ```


lochel:
Please be aware that the nightly builds are currently broken, preventing us from uploading new packages.

> Exception: Downloading artifact failed for https://build.openmodelica.org/omsimulator/nightly/linux-amd64/OMSimulator-linux-amd64-v2.1.1.post248-gf7c50c9.tar.gz

mahge (commented):
I think a top level Makefile is perhaps not a good idea. It will be confusing because one would assume it will be able to build the whole project. 

A prewritten Makefile also spoils the flexibility we got from CMake in the sense that users will not be able to chose their build program as they want. 

If possible, we should do this in the CMake files (in the respective directory) as a custom target so that it works with other build tools like Ninja, MSBuild.

lochel:
@mahge We can even remove the Makefile - it doesn't do much except for executing a python script.

mahge:
> @mahge We can even remove the Makefile - it doesn't do much except for executing a python script.

Then we can probably do it in the existing `src/pip/CMakeLists.txt`.

lochel:
Sounds good 😊 any suggestions on how to do it?

mahge:
> Sounds good 😊 any suggestions on how to do it?

Of course 🙂.  Something like this in the `src/pip/CMakeLists.txt` should be enough to create the packages. We can test it and adjust it as needed. 

```cmake
add_custom_target(omsimulator-pip-package
                  DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/setup.py
                  COMMAND python3 setup.py sdist
                  COMMENT "Creating OMSimulator pip package."
                  )
```

This will create a target named `omsimulator-pip-package`. One can then, for example, run `make omsimulator-pip-package` in the build directory to create the packages.

PR #1277, opened 2023-12-05
Allow to set CVODE parameters for ME FMUs

gabibbo97:
### Purpose

While simulating particularly stiff problems, the default settings for ME FMUs might fail with errors like

```
[CVODE ERROR]  CVode
  At t = 3e-12 and h = 1e-12, the error test failed repeatedly or with |h| = hmin.
```
or
```
[CVODE ERROR]  CVode
  At t = 1423.64, mxstep steps taken before reaching tout.

error:   [doStep] SUNDIALS_ERROR: CVode() failed with flag = -1
```

This is due to the CVODE maximum steps / NLS iterations being hardcoded into OMSimulator with fixed values.

### Approach

By adding more CLI flags, we can now customize the behavior of CVODE (I left out Adams vs BDF formula choice because if someone is trying to fiddle with these knobs he's already trying to handle a stiff problem).
Without these flags, OMSimulator should use exactly the same values as before, so the behavior should be unchanged.

### DIFF
--- a/src/OMSimulatorLib/Flags.cpp
+++ b/src/OMSimulatorLib/Flags.cpp
@@ -65,6 +65,10 @@ void oms::Flags::setDefaults()
 {
   addParametersToCSV = false;
   algLoopSolver = oms_alg_solver_kinsol;
+  cvodeMaxErrTestFails = 100;
+  cvodeMaxNLSFails = 100;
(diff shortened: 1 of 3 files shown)


lochel (approved):
Thanks for the contribution 👍 LGTM

PR #1298, opened 2024-05-15
Update setup.py

vruge:
* fix untrusted SSL certificate `https://build.openmodelica.org/omsimulator`

### Related Issues

* untrusted SSL certificate `https://build.openmodelica.org/omsimulator`

### Purpose

* ignore check

### DIFF
--- a/src/pip/setup.py
+++ b/src/pip/setup.py
@@ -70,19 +70,19 @@ class my_build_py(build_py):
 
     # download the zip directory from url
     if (sysconfig.get_platform() == 'linux-x86_64'):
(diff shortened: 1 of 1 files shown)


mziefuss:
Very useful for us, thx!

arun3688:
@vruge thanks for the fix

vruge:
I don't know if you would like to add an option like os.env["verify"] here so that you can enable verification. The error occurred in connection with a corporate network that did not accept the certificates. 

@arun3688  Thanks for the quick merge.

PR #1300, opened 2024-05-18
Drop Werror to fix gcc14 build

jschueller:
Closes https://github.com/OpenModelica/OpenModelica/issues/12432


bilderbuchi:
Just to surface that here, too: I don't think this should be merged, see https://github.com/OpenModelica/OpenModelica/issues/12432#issuecomment-2121886442

jschueller:
ok I opened https://github.com/OpenModelica/OMSimulator/issues/1301 instead

PR #1344, opened 2024-11-12
Add 'asan' option to test case headers for selective ASan testing

lochel:
As mentioned in #917, disabling ASan entirely is not ideal. However, running the full test suite with ASan enabled takes too many resources. This pull request adds a new `asan` option to the test case header, allowing us to specify specific tests to run with ASan, rather than enabling it across the entire suite.

Example usage:

```plaintext
-- status: correct
-- teardown_command: rm -rf addExternalResources1_lua/
-- linux: yes
-- ucrt64: yes
-- win: yes
-- mac: no
-- asan: yes
```

This change helps us identify memory issues without the overhead of testing the entire suite with ASan.


lochel:
@adrpo Thanks a lot!

PR #1383, opened 2024-11-27
Fixed time handling for model exchange

lochel:


src/OMSimulatorLib/SystemSC.cpp
```diff
@@ -806,15 +833,27 @@ oms_status_enu_t oms::SystemSC::doStepCVODE()
...
-      if (fmi2OK != fmistatus) logError_FMUCall("fmi2_setTime", fmus[i]);
     }
 
     if (flag == CV_ROOT_RETURN || time == tnext)
     {
-      //logInfo("event found!!! " + std::to_string(time));
+      logDebug("event found!!! " + std::to_string(time));
+
+      // set time
+      for (const auto& component : getComponents())
```
bilderbuchi, line 843: Aren't you concerned about duplicating the same code block 4 times and then having to successfully maintain this/fix bugs in this section, in the future? As opposed to, say, having this snippet in a small function called in 4 places?

lochel: I completely agree! 😊

lochel: @bilderbuchi feel free to comment on #1386

bilderbuchi: ❤️ great!

PR #1384, opened 2024-11-27
Try to fix runtest.pl

lochel:
This pull request addresses an issue in runtest.pl where $time is sometimes assigned an empty string, leading to the following error:

```
[QuarterCarModel.DisplacementDisplacement.lua:]Argument "" isn't numeric in numeric lt (<) at runtest.pl line 363
```

I am unsure why $time is empty in this context. @perost Could you provide insights or suggestions to fix this properly?

### DIFF
--- a/testsuite/partest/runtest.pl
+++ b/testsuite/partest/runtest.pl
@@ -360,7 +360,9 @@ exit_sandbox() if $sandbox_needed;
 if ($exit_status == 0) {
   exit 0;
 } else {
-  if ($time < 1) {
+  if ($time eq "") {
+    exit 1;
+  } elsif ($time < 1) {
     exit 1;
   } elsif ($time > 100) {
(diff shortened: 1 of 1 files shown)


perost:
`$time` seems to only be set here:
https://github.com/OpenModelica/OpenModelica/blob/ad594c18651234cc0709c9ad5192c68de1bd4ed0/testsuite/partest/runtest.pl#L251-L267
It might be better to use `(\d+)` instead of `(\d*)` in the regular expressions to make sure they match at least one digit, otherwise I guess it assigns an empty string to `$time` in case the time is missing in the output for some reason. I don't know why that would happen though, I've never seen this issue before.

lochel:
Not sure, but the problem seems to be in rtest, which has sometimes an uninitialized value of end_t.

PR #1445, opened 2025-03-28
set parameter values

arun3688:
### Purpose

Set parameter values on components

### DIFF
--- a/src/OMSimulatorPython/component.py
+++ b/src/OMSimulatorPython/component.py
@@ -56,7 +56,7 @@ class Component:
       for key, resources in self.parameterResources.items():
         resources.exportToSSD(component_node)
 
-  def setValue(self, cref:str, value, unit : str = None):
+  def setValue(self, cref:str, value, unit = None):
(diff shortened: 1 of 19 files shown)


src/OMSimulatorPython/component.py
```diff
@@ -1,15 +1,16 @@
...
 from OMSimulator.cref import CRef
 from OMSimulator.values import Values
-
+from OMSimulator.variable import Variable
 from OMSimulator import namespace
 
 
 class Component:
-  def __init__(self, name: CRef, fmuPath: str, connectors = list()):
+  def __init__(self, name: CRef, fmuPath: str, connectors = list(), unitDefintions = list()):
```
lochel, line 9: Using a mutable default argument such as a list can lead to unexpected behavior. If the list is modified in one instance of the class, the changes will persist across all instances of the class that use the default value. This is because the default value is evaluated only once at the time the function is defined, not each time the function is called.

lochel (approved):

PR #1466, opened 2025-04-10
export ssd:Annotations

arun3688:
### Purpose

export and import ssd:Annotations which defines the solver setting


src/OMSimulatorPython/system.py
```diff
@@ -89,6 +89,7 @@ def __init__(self, name : str, model=None):
     self.model = model
     self.elementgeometry = None
     self.systemgeometry = None
+    self.solver = None
```
lochel, line 92: self.solvers : list | None = None

arun3688: @lochel this is fixed with https://github.com/OpenModelica/OMSimulator/pull/1466/commits/147c76465cbd377deab10d9059ce35c9345a00f6

PR #1471, opened 2025-04-25
Add addConnection

lochel:


src/OMSimulatorPython/system.py
```diff
@@ -202,6 +202,34 @@ def addSSV(self, cref: CRef, resource: str):
...
+  def _addConnection(self, cref1: CRef, cref2: CRef):
+    first1 = cref1.first()
+    first2 = cref2.first()
+
+    # Both CRefs start in the same subsystem
+    if first1 == first2:
+      if first1 not in self.elements:
+        raise ValueError(f"System '{first1}' not found in '{self.name}'")
+      if first2 not in self.elements:
+        raise ValueError(f"System '{first2}' not found in '{self.name}'")
```
lochel, line 220: first1 == first2

src/OMSimulatorPython/system.py
```diff
@@ -202,6 +202,34 @@ def addSSV(self, cref: CRef, resource: str):
...
+
+      (start_element, start_connector) = self.split_cref(cref1.pop_first())
+      (end_element, end_connector) = self.split_cref(cref2.pop_first())
+      # Add the connection inside the subsystem
+      self.elements[first1].addConnection(start_element, start_connector, end_element, end_connector)
```
lochel, line 225: This will not work if the connection should be put into a deeper-nested sub-system. This function should be recursive.

src/OMSimulatorPython/system.py
```diff
@@ -202,6 +202,31 @@ def addSSV(self, cref: CRef, resource: str):
         raise ValueError(f"Component '{first}' not found in {self.name}")
       self.elements[first].addSSV(resource)
 
+  def split_cref(self, cref: CRef):
+    if cref.is_root():
+      return ("", str(cref))
+    else:
+      return (cref.first(), cref.last())
```
lochel, line 209: This doesn't work with crefs that have more than twp parts, e.g. XX.YY.ZZ. YY would just be dropped.

Also, the method shoul dbe part of the CRef class.

PR #1478, opened 2025-05-15
activate and duplicate variants

arun3688:
### Purpose

Activate and duplicate variants in ssp


src/OMSimulatorPython/component.py
```diff
@@ -18,6 +17,7 @@ def __init__(self, name: CRef, fmuType : str, fmuPath: Path | str, connectors=No
     self.value = Values()
     self.solver = None
     self.parameterResources = []
+    self.fmuType = None
```
lochel, line 20: name it implementation - to match ssp standard

PR #1491, opened 2025-09-02
Change table component MIME type to "text/csv" 

### LINKED ISSUE #1473: Portable support (OMSimulator) for MA_CSV
It is currently possible to specify CSV files, according to a sub-set of (floats) the MA format to be released. However, the resulting SSP files are not portable due to non-standard required mime types in the component. 

There is a need to export/import/simulate SSP including CSV in a portable way.

@arun3688 @lochel

AxelMatstoms:
### Related Issues
This PR resolves #1473 

### Purpose
The purpose of this PR is to support a subset of the proposed MA CSV standard for stimuli data.

### Approach
With this PR, the "correct" MIME type is used for export. Both MIME types ("legacy" application/table, and text/csv) are supported for import, which preserves compatibility with SSPs exported in previous versions of OMSimulator.

Import of SSPs with both MIME types has been tested and work as expected.


lochel:
Thanks!

PR #1503, opened 2025-10-03
fix pip setup files according to new standard

arun3688:
### Purpose

Standardize the pip setup to align with modern Python versions


lochel (approved):
Looks good 👍

PR #1504, opened 2025-10-06
Fix logging of unsupported version

phannebohm:
Adding 'fmiVersion_t' to a string does not append to the string.

Maybe this can be made prettier in the future, but at least this fixes the error.


phannebohm:
@arun3688 can you have a look and merge this?

Issue #1329, opened 2024-10-30
Time step size values from command line are ignored


rruusu:
### Description

At least for strongly coupled SSP systems, the time step size controls provided on the command line are ignore in favor of the values stored in the `oms:VariableStepSolver` annotation element in the _SystemStructure.ssd_ document.

### Expected behavior

The expected behavior is that values provided on the command line should override the values given in the SSD annotations.

### Version and OS

- Version: OMSimulator v2.1.2.post5-gce342b6-mingw-notlm-debug
- OS: Windows 11, 64-bit

Issue #1328, opened 2024-10-30
Tolerance values in strongly coupled systems in SSP are ignored


rruusu:
### Description

Tolerance values provided in the `oms:VariableStepSolver` annotation elements in SSP are entirely ignored.

The assumed expected behavior is that unless a tolerance value is provided on the command line, the solver for a strongly coupled system uses the tolerance values that are provided in the SSP annotation elements.

However, unless a command line option is provided, both the absolute and relative tolerance values are overwritten with the default value of 1e-4, instead of using the values from the `SystemStructure.ssd` document.

### Version and OS

- Version: OMSimulator v2.1.2.post5-gce342b6-mingw-notlm-debug
- OS: Not relevant

rruusu:
I will push a fix to this to my fork shortly.

PR #1520, opened 2025-10-21
Make command line options properly override options from other sources

rruusu:
### Related Issues

- #1328
- #1329

### Purpose

This commit changes the storage of command line options so that it is knowable, whether the option was provided at all.

It also changes the way the values form the flags are applied to single FMU and SSP simulations, so that the default values for flags are only used if the corresponding value is not defined by the FMU or SSP, and all values in FMUs or SSPs can be overridden from the flags.

### Approach

The flag values are stored in a class `FlagValue` that stores both the value and a boolean variable `given` that defines whether the flag value was actually provided by the user.

The value of the `given` variable is checked to decide whether the values from FMUs or SSPs should be overridden.

### DIFF
--- a/src/OMSimulatorLib/Flags.cpp
+++ b/src/OMSimulatorLib/Flags.cpp
@@ -45,10 +45,47 @@
 #include <regex>
 #include <sstream>
 
-oms::Flags::Flags()
+oms::Flags::FlagValues::FlagValues() :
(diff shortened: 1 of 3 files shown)


rruusu:
This pull request was made obsolete by merge of #1521.

Issue #1330, opened 2024-10-30
Option for emitting all integrator time steps


rruusu:
### Current status

Currently OMSimulator only emits data into the result file using a fixed time step size, either based on the maximum time step size or the number of intervals provided by the `--intervals` option.

### Feature description

It is sometimes useful to get output that shows all steps made by the variable step solver, especially for debugging purposes.

One option for expressing this to OMSimulator could be by setting `--intervals=0` on the command line.

PR #1522, opened 2025-10-21
Output all CVODE steps

rruusu:
These changes are cherry picked from the branch ["work"](https://github.com/rruusu/OMSimulator/tree/work), which includes changes originally made to the master branch in October 2024. I hadn't had the time earlier to rebase my code on the major refactoring done to SystemSC.cpp on 19.11.2024 (6d7a397).

Note: This request is based on the **maintenance/v2.1** branch, on which I have been working on.

### Related Issues

- #1330
- #1530

### Purpose

This pull request changes `SystemSC` so that:
- `doStep()` always advances a single step of CVODE.
- `stepUntil(double)` steps to exactly the given time value.
- Results are emitted for the nearest CVODE step according to the defined output interval.
  - This allows emitting results at a faster rate than the maximum step size.

### Approach

`doStepCVODE()` and `doStepEuler()` are changed to get the end time from `stepUntil()`, instead of just using the stopping time for the whole model.

`doStepCVODE()` is changed to:
- Return after each CVODE time step
  - Single call from `doStep()`, calls in a loop from `stepUntil()`.
  - This allows the progress bar to be updated during simulation.
- CVODE is called with mode `CV_ONE_STEP` instead of `CV_NORMAL`.
  - If the single step goes beyond the end time, a second call with `CV_NORMAL` is made.
  - The second call only interpolates the values, instead of advancing the solver.

Also eliminates unnecessary calls to `getContinuousStates()` after the event loop but before updating inputs.

Note, that this change makes a logging interval of zero emit **all solver time steps** instead using the maximum step size. This may **drastically increase** the output file size, if the actual time step is much smaller than the maximum step size. The remedy is to set the logging interval to the maximum step size.

Note: The indentation of the while loop in `doStepCVODE` is unchanged to create simpler view of the changes.

### DIFF
--- a/src/OMSimulatorLib/AlgLoop.cpp
+++ b/src/OMSimulatorLib/AlgLoop.cpp
@@ -53,7 +53,7 @@ inline bool checkFlag(int flag, std::string functionName)
     logError("SUNDIALS_ERROR: " + functionName + " failed with flag = " + std::to_string(flag));
     return false;
   }
-  logDebug("SUNDIALS_INFO: " + functionName + " failed with flag = " + std::to_string(flag));
+  logDebug("SUNDIALS_INFO: " + functionName + " succeeded with flag = " + std::to_string(flag));
   return true;
 }
 
@@ -391,6 +391,12 @@ oms::KinsolSolver* oms::KinsolSolver::NewKinsolSolver(const int algLoopNum, cons
   return kinsolSolver;
 }
 
+static std::string to_string(double value) {
+  std::ostringstream s;
+  s << value;
+  return s.str();
+}
+
 /**
  * @brief Solve algebraic system with KINSOL
  *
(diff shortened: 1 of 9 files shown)


rruusu:
The approach in dd04f813926811c7a624f52223a852ea4af68889 is clumsy, but I couldn't figure out any other way to force CVODE to always stop at or before `tnext` in the presence of roots, while keeping an internal state that is consistent with the state of the system.

The issue was that using a second call with `CV_NORMAL` does not correctly return the state at `tnext`, when the previous call with `CV_ONE_STEP` has found a root at a time after `tnext`.

This approach results in skipping the output of possible intermediate time steps, when the stopping time is at most maximum step size away from the current time.

It is interesting that CVODE returns slightly different state values depending on whether the root is found in a call made using `CV_NORMAL` or `CV_ONE_STEP`. `CV_ONE_STEP` apparently produces a smaller numerical error, at least in the case of *EventTest.lua*. (See decimal changes in baseline.)

PR #1587, opened 2026-05-13
Add setString to procedural API

AxelMatstoms:
### Related Issues
setString is missing from the procedural Python API.

### Purpose
To enable setting string parameters using the procedural API, add the missing setString method to the procedural API. The method was already implemented in the object oriented API and implemented in the capi wrapper.

### Approach
```diff
+    self.setString = Scope._capi.setString
```


AnHeuermann:
This PR is targeting the maintenance branch `OpenModelica:maintenance/v2.1` instead of `master`. Seems to be intended, since the old API was removed in 3420839f.

Issue #1541, opened 2025-11-18
Update build-deps Dockerfiles


AnHeuermann:
## Issue

The CI always uses apt to install additional build dependencies like asan, see e.g. https://github.com/OpenModelica/OMSimulator/blob/master/.CI/cache/Dockerfile.
`apt` doesn't have a stable CLI and always installing packages slows down the build..

Instead the CI should use pre-build Docker image to speed up test execution time.
In addition there are multiple very old Docker iamges that need to be updated: 

- alpine (5 years)
- cache-32 (2 years)
- cache-arm32 (7 years)

## Tasks

- [ ] Create new branches for 32bit and ARM in https://github.com/OpenModelica/build-deps with OMSimulator specific dependencies.
- [ ] Update build-deps images in `.CI/`.

AnHeuermann:
We could add a build-deps image for 32 ARM based on some Debian release. Not sure if really needed right now, so let's skip that one.

PR #1594, opened 2026-07-07
Modernize CI Docker images and add dev containers

AnHeuermann:
### Related Issues

Building Dockerfiles takes time and fails a lot.
Partial fix for https://github.com/OpenModelica/OMSimulator/issues/1541.

### Purpose

Moves the Linux CI off the locally-built `.CI/cache*` Dockerfiles and onto the new OS-keyed `build-deps` images, refreshes the Alpine image, and adds VS Code dev containers that reproduce the CI build environments.

### Approach

- Replace the `.CI/cache` `dockerfile { }` agents with prebuilt, images from `docker.openmodelica.org/build-deps`:
  - `linux64` → **`linux64-noble`** on `build-deps:ubuntu-24.04-omsimulator`
  - `linux64-jammy` now uses `build-deps:ubuntu-22.04` (was the pinned `ghcr.io/openmodelica/build-deps:v1.22.3`)
  - `linux64-asan` → **`linux64-resolute-asan`** on `build-deps:ubuntu-26.04`
  - `/cache/runtest` volume mounts and the ASAN `--cap-add SYS_PTRACE` `--privileged` / memory-limit args are preserved; `additionalBuildArgs '--pull'` becomes `alwaysPull true`.
- Refactor `buildOMS()` to take additional CMake arguments as an input (`buildOMS(extraCMakeArgs='')`), threaded through the Linux, macOS and Windows branches.
  - Pass AddressSanitizer as a CMake flag directly — `buildOMS('-DASAN=ON')` — instead of the `ASAN` environment variable, and drop the now-unused `ASAN` env from the build stage (it is still set in the test stage, which `partest()` needs).
- Update `.CI/alpine` from the EOL `alpine:3.12` to `alpine:3.24` (latest stable; gcc 15.2, CMake 4.2). Verified by building OMSimulator in the image.
- Remove the dead, unreferenced `.CI/cache`, `.CI/cache-32` and `.CI/cache-arm32` Dockerfiles (32-bit x86 / arm32 are no longer built; i386 Ubuntu no longer exists upstream).
- Add `.devcontainer/` with containers mirroring the CI build environments:


AnHeuermann:
Waiting for https://github.com/OpenModelica/build-deps to build and upload rolling images. Looks like GitHub has some trouble right now.

Issue #1609, opened 2026-07-29
Events are not detected for FMI 3.0 co-simulation


AnHeuermann:
## Description

OMSimulator never asks an FMI 3.0 co-simulation FMU to report its events, so all
events that happen inside `fmi3DoStep` are lost. The result file only contains
the communication points and the discontinuity is smeared over one step instead
of being stored as two data points.

`Stair3.fmu` declares everything that is needed for this to work:

```xml
<CoSimulation ... canHandleVariableCommunicationStepSize="true"
                  providesIntermediateUpdate="true"
                  canReturnEarlyAfterIntermediateUpdate="true"
                  hasEventMode="true" />
```

but OMSimulator opts out of event mode at instantiation and ignores the event
flags returned by `fmi3DoStep`, see [What is missing](#what-is-missing) below.

This is a follow-up to #1603. The model-exchange side is fixed there; this issue
is only about co-simulation. Note that FMI 2.0 co-simulation cannot be fixed —
it has no event mode, no early return and no way to observe an event inside
`fmi2DoStep` — so this is specific to FMI 3.0.

## Steps to reproduce the behavior

```bash
$ cd /path/to/OMSimulator/testsuite/reference-fmus/3.0
$ python3 Stair.py
Loading FMI version 3...
info:    Result file: Stair-cs3.mat (bufferSize=10)
error:   [compareSeries] ResultReader::compareSeries: different number of events: 9 != 0
signal counter is not equal
```

The reference `testsuite/references/Stair-cs.mat` holds 9 events (the counter
increments at t = 1, 2, ..., 9), the result file holds none.

Inspecting `Stair-cs3.mat` shows the plain communication grid, no data points
sharing a time stamp:

```
t       : 0.0  0.2  0.4  0.6  0.8  1.0  1.2  ...
counter : 1    1    1    1    1    2    2    ...
```

## Expected behavior

At every event the result file should contain two data points with the same
time stamp, holding the value before and after the event:

```
t       : ...  0.8  1.0  1.0  1.2  ...
counter : ...  1    1    2    2    ...
```

i.e. the same trajectory the model-exchange run already produces after #1603.

## What is missing

1. **Event mode is not requested at instantiation.**
   [`ComponentFMU3CS::instantiate()`](../src/OMSimulatorLib/ComponentFMU3CS.cpp#L646)
   hardcodes both `eventModeUsed` and `earlyReturnAllowed` to `fmi3False`:

   ```cpp
[...]

PR #1610, opened 2026-07-29
Store the rhs of an event in the result file

### LINKED ISSUE #1557: Duplicate lines in CSV output for CS FMU
### Description

There are 2 identical lines written to the CSV output for each timestamp when simulating a cosimulation FMU. If the FMU is built as a model exchange FMU, this does not occur.

Create the files `BouncingBall.mo` and `build.mos`, then run the shell commands, and inspect `res.csv`.

##### `BouncingBall.mo`
```
model BouncingBall "The 'classic' bouncing ball model"
  type Height=Real(unit="m");
  type Velocity=Real(unit="m/s");
  parameter Real e=0.8 "Coefficient of restitution";
  parameter Height h0=1.0 "Initial height";
  Height h "Height";
  Velocity v(start=0.0, fixed=true) "Velocity";
initial equation
  h = h0;
equation
  v = der(h);
  der(v) = -9.81;
  when h<0 then
    reinit(v, -e*pre(v));
  end when;
end BouncingBall;
```

##### `build.mos`
```
loadFile("BouncingBall.mo");
buildModelFMU(BouncingBall, version="2.0", fmuType="cs");
```

##### Shell commands
```sh
omc build.mos
OMSimulator --resultFile=res.csv BouncingBall.fmu
```

##### Output (`res.csv`)

```csv
time,h,v,der(h),der(v)
0, 1, 0, 0, -9.81
0.002, 1, -0.01962, -0.01962, -9.81
0.002, 1, -0.01962, -0.01962, -9.81
0.004, 0.99996076, -0.03924, -0.03924, -9.81
0.004, 0.99996076, -0.03924, -0.03924, -9.81
0.006, 0.99988228, -0.05886, -0.05886, -9.81
0.006, 0.99988228, -0.05886, -0.05886, -9.81
0.008, 0.99976456, -0.07848, -0.07848, -9.81
0.008, 0.99976456, -0.07848, -0.07848, -9.81
```

### Expected behavior

I expect the output file to not have 2 entries for each timestamp.

### Screenshots

N/A

### Version and OS

- Version: OMSimulator 2.1.3~9-ge3a3aeb4-linux-notlm-debug
- OMC Version: OpenModelica 1.26.1
- OS: Debian GNU/Linux 13 (trixie)

### Extra details

As far as I can tell this is an issue with OMSimulator and not the FMU itself, because fmusim from [modelica/Reference-FMUs](http://github.com/modelica/Reference-FMUs) doesn't exhibit this behavior.

### LINKED ISSUE #1603: Events not saved correctly in result file
## Description

OMSimulator isn't finding the events in Stair.fmu with enough precision.

Image

## Steps to reproduce the behavior

Run

```bash
$ cd /path/to/OMSimulator/testsuite/resources
$ /path/to/OMSimulator/build/install_cmake/bin/OMSimulator -m=me Stair.fmu
info:    Set temp directory to    "/home/andreas/workdir/OM/OMSimulator/testsuite/resources"
info:    Set working directory to "/home/andreas/workdir/OM/OMSimulator/testsuite/resources"
info:    New model "Stair" with corresponding temp directory "/home/andreas/workdir/OM/OMSimulator/testsuite/resources/Stair-2kz4f13y"
info:    Set working directory to "/home/andreas/workdir/OM/OMSimulator/testsuite/resources"
info:    *** FMU Simulation Info ***
          - model:     Stair (model-exchange)
          - path:      Stair.fmu
          - startTime: 0.000000
          - stopTime:  10.000000
          - tolerance: 0.000100
          - stepSize:  0.200000
info:    model doesn't contain any continuous state
info:    maximum step size for 'Stair.root': 0.200000
info:    Result file: Stair_res.mat (bufferSize=10)
info:    Simulation terminated by FMU Stair.root.Stair at time 9.000000
info:    Final Statistics for 'Stair.root':
         NumSteps = 0 NumRhsEvals  = 0 NumLinSolvSetups = 0
         NumNonlinSolvIters = 0 NumNonlinSolvConvFails = 0 NumErrTestFails = 0
```

## Expected behavior

Find events at time = 1,2,3,... and save two data points in result file resulting in "jump" at every event.

## Version and OS

- Version: OMSimulator v3.0.0.post154-g2d2b004-linux-debug
- OS: Ubuntu 26.04

### LINKED ISSUE #1606: Reference result files missing events
## Issue

The reference result files in `testsuite/references` are missing the events (indicated by two values at one time point) in the MAT files.

Why are there two versions of the reference results for model exchange and co-simulation? The reference shouldn't change, both versions need to be the same.
If Co-Simulation produces slightly different results (e.g. more time points from the master algorithm) the `compareSimulationResults` needs to handle these.

## Task

* [ ] Update Stair-cs.mat: Add missing events
* [x] Update Stair-me.mat: Add missing events
* [ ] Update BouncingBall-cs.mat: Add missing events
* [x] Update BouncingBall-me.mat: Add missing events
* [ ] Update Dahlquist-cs: Remove "fake" events at every 0.1s. The FMU does internal steps with step size 0.1, the master algorithm does steps with step size 0.001 resulting in two values at one time point
* [x] RealNetwork1.mat: Remove (unused)
* [x] SSC.mat: Remove (unused)

AnHeuermann:
## Related Issues

Model Exchange part for https://github.com/OpenModelica/OMSimulator/issues/1603 and https://github.com/OpenModelica/OMSimulator/issues/1606.
Fixes https://github.com/OpenModelica/OMSimulator/issues/1557.

## Purpose

`emit()` bailed out whenever the time stamp had already been used, so the rhs of an event was never stored.
Let a forced emit pass that check, so model exchange stores the value before and after each event.

## Approach

* Make `emit` honor `force` input
* Regenerates the ME references for BouncingBall and Stair.


AnHeuermann:
New reference file and the solution for Stair.fmu have sharp events, including the final event at stop time:

grafik

AnHeuermann:
Still too many emits in the CSV file. For Stair-me.py (changing the output to CSV) I get one emit too many:

```csv
"time","default.Stair.time","default.Stair.counter"
[...]
2, 2, 2
2, 2, 2
2, 2, 3
[...]
```

There is an event at `time=2.0`, but only one. My assumption is that this is caused by the equality checks on floating point numbers.

Update: Root cause is that we save the first data point at `1.9999999999999998` and for CSV files this is written to file with rounding `%.12g`. So that are different time points, but this get's lost after rounding.

AnHeuermann:
@arun3688 looks like we are removing and adding the same thing in https://github.com/OpenModelica/OMSimulator/pull/1580 and here.

There are a few things we should do to get this sorted once and for all:

* Update `compareSeries` to also check events for testing (equal number of events, event times approx equal)
* Ensure the reference files match the expected output from the reference FMU documentation
  * Stairs.fmu needs to have **exactly** 10 events (9 for the counter, 1 final termination event), see https://github.com/OpenModelica/OMSimulator/blob/master/testsuite/resources/Stair/documentation/result.svg. OMSimulator needs to detect all of them and save them in the MAT/CSV file correctly.

AnHeuermann:
This PR should be ready and fix most of the event issues we have including 2.0 Co-Simulation FMUs.
It doesn't cover the 3.0 Co-Simulation issue from https://github.com/OpenModelica/OMSimulator/issues/1609 and the test for Stair.fmu *should* fail.

AnHeuermann:
After this is merged we should update OpenModelica's pointer to OMSimulator sub-module. This will change the behavior for some of the test we use OMSimulator for (e.g. adding the rhs of events at stop time).

arun3688 (approved):
@AnHeuermann Looks good

PR #1614, opened 2026-07-30
Replace the Makefile testsuite with CTest

### LINKED ISSUE #1612: Switch to CTest
Move from the Makefile tests powered by the perl script `rtest` to CTest.

Upsides

* Remove Makefiles from CMake project
* Separate test driver and test detection
* Prevent source tree polution
* Parallelism for free

AnHeuermann:
### Related Issues

Fixes #1612.

### Purpose

* Get rid of Makefiles in CMake project
* Improve parallelization and isolation
* Keep source directory clean

### Approach

Discovery, parallelization and reporting are handled by CTest now, so rtest and partest are gone. What is left of them is testsuite/runtest.py, which runs a single test file and compares its output against the expected result embedded in the file, using omc-diff as before.

Every test runs in its own working directory, so tests no longer write into the source tree and teardown_command (plus the unused setup_command) could be dropped. The test files moved to testsuite/tests/. Jenkins runs ctest --output-junit instead of partest.


AnHeuermann:
@coderabbitai review

PR #1617, opened 2026-07-30
Re-enabling ASAN tests

### LINKED ISSUE #1615: Fix ASAN tests and memory  leaks
### Description

There isn't a single test in the testsuite with `asan: yes`, so no address sanitizer test is ever performed. This results in a useless Jenkins stage for ASAN.

Add `## asan: yes` to a few tests and see how much memory we leak.
For example for `testsuite/tests/reference-fmus/2.0/me/Feedthrough-me.py`

```
=================================================================
==225087==ERROR: LeakSanitizer: detected memory leaks
[...]
SUMMARY: AddressSanitizer: 132971 byte(s) leaked in 160 allocation(s).
```

### Expected behavior

* Test ASAN
* Don't leak memory

### Version and OS

- Version: OMSimulator v3.0.0.post157-gfcad4fe-linux-debug
- OS: [e.g. Win10, 64bit]

AnHeuermann:
## Related Issues

Fixes https://github.com/OpenModelica/OMSimulator/issues/1615.

## Purpose

* Enable ASAN again
* Updated testsuite to enable ASAN everywhere possible
  * Fix memory issues that popped up while it was disabled or never discoverd
  * Disabled ASAN tests still leak memory
* Update Build configuration to be more strict on all used compiler

### Bugs fixed in OMSimulator

**Heap-buffer-overflow in `SystemSC`** (b917d7b7)

`callEventUpdate` and `terminateSimulation` were allocated in a default member initializer, `new bool[fmus.size()]()`. That runs in the constructor, where `fmus` is still empty — `initialize()` only fills it later — so both arrays were always zero length. Every ME step then wrote one flag per FMU past the end of them. The arrays are now `std::unique_ptr<bool[]>` sized in `initialize()` once `fmus` is known. `unique_ptr` also closes two related problems: `terminate()` `delete[]`-ed the arrays without clearing the pointers while `initialize()` never re-allocated them (dangling pointer on re-initialize), and `~SystemSC()` freed nothing at all (leak when `terminate()` is never called).

**KINSOL solver leaked per algebraic loop** (01dcf3fd)

`AlgLoop` owned `KinsolSolver*` as a raw pointer and never deleted it, leaking the KINSOL memory block, five `N_Vector`s, the dense matrix, the linear solver and the user-data struct for every algebraic loop. `kinsolData` was also left uninitialized when the fixed-point solver was selected, and the ~12 error returns in `NewKinsolSolver()` leaked the half-built object. Now a `unique_ptr` member, with the factory holding one until the object is complete.

### Build configuration

- ASAN now works with Clang, not only GCC (4e5e02f5). The flags used to sit  inside a GNU-only branch, so `-DASAN=ON` on a Clang build silently produced an uninstrumented library — it still reported leaks through the preloaded libasan
  interceptors, but detected no overflow or use-after-free at all. A `FATAL_ERROR` now rejects compilers too old to support the sanitizer.
- MSVC equivalents of the `-Werror=` memory warnings (2193adb7): C4700, C4701, C4703, C4715, C4716 and C4172 are raised to the default warning level and turned into errors.

### Testsuite

- New `testsuite/lsan-suppressions.txt`, wired into `runtest.py` via `LSAN_OPTIONS`. It covers only non-OMSimulator modules: the uutils coreutils processes the `OMSimulatorPython3` wrapper starts, CPython's own interpreter state, and lxml/libxml2/libxslt. `print_suppressions=0` keeps the reports out of the compared test output. The file documents why matching on module paths is safe here and why `fast_unwind_on_malloc=0` must not be set.
- `asan: yes` on 82 test cases, `asan: no` with an explanatory line on 23 (774f9037, ab681d02) — see the limitation below.
- Dropped the unused `from numpy import uint64` from the two `FeedthroughSetValue` tests and `python3-numpy` from the coverage workflow (8cfb221a). numpy leaks ~683 bytes registering its ufuncs at import, in stacks too shallow to suppress by interpreter frame.

### Known limitation

23 tests are marked `asan: no`. All of them fail on leaks — and in the two `embrace` cases a use-after-free — that occur *inside the FMU binaries*, not in OMSimulator: OpenModelica-generated FMUs leak in `<Model>_function_initSynchronous`, `FMU_16_aero_interface.so` leaks ~82 kB during `fmi2Instantiate`, and Dymola's
`ECS_HW.so` reads the pointer returned by `setlocale()` after a later call has freed it. All of these tests do call `terminate()` and `delete()`, so `fmi2FreeInstance` is issued correctly. Suppressions cannot help: once the FMU is
`dlclose`d the stacks are `#0 malloc / #1 <unknown module>` with nothing left to
match on. Each disabled test carries a one-line comment naming the reason.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -25,12 +25,11 @@ jobs:
           submodules: recursive
 
       # Everything else OMSimulator needs is vendored in 3rdParty/. flex is for
-      # omc-diff, which the testsuite compares result files with, and numpy for
-      # the CompositeModels tests that import it.
+      # omc-diff, which the testsuite compares result files with.
       - name: Install dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y cmake g++ flex python3 python3-numpy gcovr ccache
+          sudo apt-get install -y cmake g++ flex python3 gcovr ccache
 
       # Only 3rdParty benefits much: the instrumented sources are rebuilt at -O0
       # anyway. It still takes the bulk of a cold build off the critical path.
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ imgui.ini
 /OMSimulator.creator.user
 /OMSimulator.files
 /OMSimulator.includes
+
+# Python files
+__pycache__/
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,6 @@ pipeline {
     booleanParam(name: 'MACOS_ARM64', defaultValue: false, description: 'Build with macOS-arm64 (M1 mac)')
     booleanParam(name: 'SUBMODULE_UPDATE', defaultValue: false, description: 'Allow pull request to update submodules (disabled by default due to common user errors)')
     booleanParam(name: 'UPLOAD_BUILD_OPENMODELICA', defaultValue: false, description: 'Upload install artifacts to build.openmodelica.org/omsimulator. Activates MINGW_UCRT64 as well.')
-    booleanParam(name: 'ASAN', defaultValue: false, description: 'Build and test with AddressSanitizer (disabled by default while the reported leaks are unfixed)')
     string(name: 'CTEST_FLAGS', defaultValue: '', description: 'Extra flags passed to ctest, e.g. -R api')
   }
   stages {
@@ -48,12 +47,6 @@ pipeline {
     stage('build-in-parallel') {
       parallel {
         stage('linux64-resolute-asan') {
-          // Disabled until the leaks AddressSanitizer reports have been fixed.
-          // Set the ASAN build parameter to run it in the meantime.
-          when {
-            expression { return params.ASAN }
-            beforeAgent true
-          }
           stages {
             stage('build-asan') {
               agent {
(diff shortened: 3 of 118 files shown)


AnHeuermann:
We have a few real memory issues:

What | Reports | Where
-- | -- | --
Leaks under fmi2_instantiate | 565 | all reachable from oms_instantiate → oms::ComponentFMUCS::instantiate() (ComponentFMUCS.cpp:631)
SUNDIALS leaks (N_VNewEmpty, N_VClone_Serial, SUNDenseMatrix, SUNLinSol_Dense) | ~50 | oms::KinsolSolver::NewKinsolSolver, under oms_initialize
heap-use-after-free | 2 | inside the FMU binary ECS_HW.so, in sscanfCx2_underscore — AircraftVehicleDemonstrator.embrace1/embrace2

AnHeuermann:
The remaining issues are basically FMUs leaking memory. We should check if updating the OpenModelica ones improves the situation.

Not sure if we can update the remaining FMUs. For those disabling ASAN might be the only option.

AnHeuermann:
Looks like I need to finally do https://github.com/OpenModelica/OpenModelica/pull/14638....
The old multiarch/crossbuild image isn't compatible with OpenModelica v1.27.

Issue #1628, opened 2026-08-19
SSM files with no mapping entries causes crash during instantiation


AxelMatstoms:
### Description
Empty SSM files seem to cause OMS to bypass some of the logic that normally allows values in SSV files with no corresponding parameters in the system. 

### Steps to reproduce the behavior
SSP file demonstrating issue. In this case we have an SSV file with some parameters with no corresponding parameter in the system.
```xml
<?xml version='1.0' encoding='utf-8'?>
<ssv:ParameterSet xmlns:ssc="http://ssp-standard.org/SSP1/SystemStructureCommon" xmlns:ssv="http://ssp-standard.org/SSP1/SystemStructureParameterValues" version="2.0" name="parameters">
  <ssv:Parameters>
    <ssv:Parameter name="t_start">
      <ssv:Real value="0.0" unit="None"/>
    </ssv:Parameter>
    <ssv:Parameter name="t_end">
      <ssv:Real value="10.0" unit="None"/>
    </ssv:Parameter>
  </ssv:Parameters>
</ssv:ParameterSet>
```
and an SSM file with no entries:
```xml
<?xml version='1.0' encoding='utf-8'?>
<ssm:ParameterMapping xmlns:ssc="http://ssp-standard.org/SSP1/SystemStructureCommon" xmlns:ssm="http://ssp-standard.org/SSP1/SystemStructureParameterMapping" version="2.0">
</ssm:ParameterMapping>
```
**SSP file:**
[example.zip](https://github.com/user-attachments/files/31222077/example.zip)

```Python
from OMSimulator import SSP

ssp = SSP("example.ssp")

instantiated_model = ssp.instantiate()

instantiated_model.setResultFile("out.csv")
instantiated_model.setStopTime(10)
instantiated_model.setFixedStepSize(0.05)

instantiated_model.initialize()
instantiated_model.simulate()
instantiated_model.terminate()
instantiated_model.delete()
```

When simulating this with latest version of OMS `OMSimulator v3.0.0.post164-g33c40b0-linux` i get the following error:
```
error:   [setReal] Unknown signal "model.root.t_start"
Traceback (most recent call last):
  File "/home/axema/Documents/oms3-ssv-example/sim.py", line 5, in <module>
    instantiated_model = ssp.instantiate()
                         ^^^^^^^^^^^^^^^^^
[...]

PR #1634, opened 2026-08-25
Change branch condition in setStartValues

AxelMatstoms:
### Related Issues
#1628 

### Purpose
The SSP standard allows empty SSM files. When an SSM is present, no values except those in the SSM should be applied. This behavior corresponds to the first branch in setStartValues. Previously the condition in this if statement was `ssm and ssm.mappingEntry`. This means empty SSM files caused the other branch to be selected. This is contrary to what the standard requires.

### Approach
 Using `ssm is not None` instead ensures compliance to the SSP standard, by using the SSM branch even if there are no mappings. The effect is that no values are applied if an SSV is referenced through an "empty" SSM, which is what the standard implies.

### DIFF
--- a/src/OMSimulatorPython/instantiated_model.py
+++ b/src/OMSimulatorPython/instantiated_model.py
@@ -351,7 +351,7 @@ class InstantiatedModel:
       raise KeyError(f"Missing required key: '{systemName}'")
 
     ## apply ssm mapping if exist and apply start values from mapped parameters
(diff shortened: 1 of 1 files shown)


arun3688:
@AxelMatstoms it is fixed with this commit https://github.com/OpenModelica/OMSimulator/commit/8ce1f2d97935abc37d850b736c74a56da25fd0b5, So i am closing this PR

Issue #1550, opened 2025-12-05
Unable to change parameter value in SSP


larsviktorlarsson:
### Description

When I change a parameter value in an .ssd file from the python api, it does not have an effect on a certain FMU.

### Steps to reproduce the behavior
My python script:

from OMSimulator import SSP as oms_ssp

from OMSimulator import CRef

from OMSimulator import Capi as oms_capi

try:
    oms_capi.terminate("model")
except OSError:
    pass
try:
    oms_capi.delete("model")
except OSError:
    pass

oms_model = oms_ssp("ParkerSimPack.ssp")
oms_model.activeVariantName = "D16_auto_stimuli"
oms_model_inst = oms_model.instantiate()

oms_model_inst.setResultFile('D16_auto_stimuli_results.csv')

oms_model_inst.initialize()

oms_model_inst.setValue(CRef("D16_auto_stimuli","interface_model.k_p"),0)

oms_model_inst.simulate()

oms_model_inst.terminate()
oms_model_inst.delete()

### Expected behavior

When I plot the results in Matlab i get:

<img width="1662" height="818" alt="Image" src="https://github.com/user-attachments/assets/715b2e6c-76a6-4363-8c44-26657e4d3dea" />

But, the variagble theta should be zero after I have changed the parameter value of k_p to 0. If I simulate in easy-ssp I get (which I expect in OMSimulator as well):

<img width="1782" height="695" alt="Image" src="https://github.com/user-attachments/assets/08629b1e-38d7-4b80-b5c2-c336cfae1bae" />

### Model
[ParkerSimPack.zip](https://github.com/user-attachments/files/23957983/ParkerSimPack.zip)

### Version and OS

- Version: 'OMSimulator v3.0.0.post103-g53bba5e-win'

- OS: Win11 64 bit

arun3688:
@larsviktorlarsson I did some debugging and i found few things, the composite model has algebraic loops, due to the following connections

```
 valve_model.qA -> interface_model.qA.out.y
 valve_model.qB -> interface_model.qB.out.y
```

And when you set the value, you have to set it before initialization, so it should be 

```
oms_model_inst = oms_model.instantiate()
oms_model_inst.setResultFile('D16_auto_stimuli_results.csv')
oms_model_inst.setValue(CRef("D16_auto_stimuli","interface_model.k_p"),0) // set it before initialize
oms_model_inst.initialize()
```
[...]

PR #1642, opened 2026-08-31
Move fmi enter initialization mode out of FMU instantiation

arun3688:
### Related Issues

https://github.com/OpenModelica/OMSimulator/issues/1550

### Purpose

Move fmi2_enterInitializationMode() from instantiate() to the system initialization phase so that FMUs enter initialization mode only when the system is initialized. This also allows setValues to be properly initialized before entering initialization mode.


arun3688:
@this is fixed with this commit 527c4fae14a01d3508545be6300a90cdadd0a5d3

